PR #37, opened 2018-05-15
Add reference implementation

justinfagnani:


package-name-map/README.md
```diff
@@ -0,0 +1,6 @@
+# package-name-map
```
domenic: I'd prefer the folder be called reference-implementation (and the h1 be "Reference implementation")

package-name-map/package.json
```diff
@@ -0,0 +1,31 @@
+{
+  "name": "package-name-map",
```
domenic: "private": true?

package-name-map/package.json
```diff
@@ -0,0 +1,31 @@
...
+  "version": "0.1.0",
```
domenic: I always start with "1.0.0-pre" myself

package-name-map/src/package-name-map.ts
```diff
@@ -0,0 +1,248 @@
...
+  private _map: Scope;
+
+  constructor(map: Scope) {
+    this._map = map;
+  }
+
+  resolve(specifier: string, referrerURL: string): string {
+    // 1. If specifier parses as absolute URL, return the specifier.
+    try {
+      new URL(specifier);
```
domenic: Since this uses global `URL`, I think the engines in package.json should be Node v10+

package-name-map/src/package-name-map.ts
```diff
@@ -0,0 +1,248 @@
...
+      return specifier;
+    } catch (e) {
+      // proceed
+    }
+
+    // 2. If specifier starts with `./`, `../`,  or `/`, return the specifier.
+    if (/^\.{0,2}\//.test(specifier)) {
+      return specifier;
```
domenic: Shouldn't this return `new URL(specifier, referrerURL).href`?

package-name-map/src/package-name-map.ts
```diff
@@ -0,0 +1,248 @@
...
+    packageName: foundPackageName,
+    package: foundPackage,
+  };
+};
+
+/**
+ * Joins together two or more strings, ensuring there's a '/' character
+ * between each.
+ */
+const joinPaths = (...paths: Array<string | undefined>) => {
```
domenic: Hmm why is `undefined` allowed in this array? (Makes it harder to write spec text for.)

package-name-map/src/package-name-map.ts
```diff
@@ -0,0 +1,248 @@
...
+      result += '/';
+    }
+    result += path;
+  }
+  return result;
+};
+
+/**
+ * Returns `true` iff `prefix` is a string prefix of `str` and `prefix` the rest
+ * of `str` are separated by a path separator ('/').
```
domenic: Missing word

package-name-map/src/test/package-name-map_test.ts
```diff
@@ -0,0 +1,226 @@
...
+          },
+          '@polymer/polymer-foo': {
+            path: '/node_modules/@polymer/polymer-foo',
+            main: 'polymer-foo.js',
+          },
+        },
+      });
+
+      test('resolves package name for a package with only a main', () => {
+        assert.equal(map.resolve('app', referrerURL), 'app/src/index.js');
```
domenic: The result should be `http://foo.com/app/src/index.js`; we want `resolve()` to always return an absolute URL.

package-name-map/src/test/package-name-map_test.ts
```diff
@@ -0,0 +1,226 @@
...
+
+import chai = require('chai');
+
+import {PackageNameMap, isPathSegmentPrefix} from '../package-name-map.js';
+
+const {assert} = chai;
+
+suite('PackageNameMap', () => {
+  suite('resolve', () => {
+    const referrerURL = 'http://foo.com/';
```
domenic: Maybe the referrerURL should be something more like `https://foo.com/js/entry.js` for a more realistic starting point.

package-name-map/src/test/package-name-map_test.ts
```diff
@@ -0,0 +1,226 @@
...
+        assert.equal(map.resolve('./foo', referrerURL), './foo');
+        assert.equal(map.resolve('../foo', referrerURL), '../foo');
+      });
+    });
+
+    suite('top-level package names', () => {
+      const map = new PackageNameMap({
+        packages: {
+          app: {
+            main: 'src/index.js',
```
domenic: This is interesting because consider `referrerURL` of `https://foo.com/js/entry.js`. What should the result be, since you are using a relative URL? Should the answer be different for `https://foo.com/js/nested/entry.js`? Should the answer depend on where the package name map itself is hosted?

It seems like either we need to disallow relative paths, or we need to turn them into absolute paths ASAP, probably using the package name map's URL (which is distinct from the referrerURL).

package-name-map/src/test/package-name-map_test.ts
```diff
@@ -0,0 +1,226 @@
...
+          '/node_modules/lodash-es/lodash.js'
+        );
+      });
+
+      test('resolves package names in outer scopes from deeper scopes', () => {
+        assert.equal(
+          map.resolve('moment', '/node_modules/lodash-es/subpackage'),
+          '/node_modules/lodash-es/node_modules/moment/moment.js'
+        );
+      });
```
domenic: Worth adding a test of not-moment here, i.e. a test of a package inside a scope inside a scope, to make sure the algorithm works beyond one level.

package-name-map/src/test/package-name-map_test.ts
```diff
@@ -0,0 +1,226 @@
...
+    });
+  });
+});
+
+suite('isPathSegmentPrefix', () => {
```
domenic: Maybe best in a separate file since it's not something we'd eventually port to web platform tests?

package-name-map/README.md
justinfagnani: If this is published as an npm package, we'd want a descriptive name, like `package-name-map`.

package-name-map/package.json
justinfagnani: I was hoping to publish this to use in other prospective libraries/tools.

package-name-map/src/package-name-map.ts
justinfagnani: already have that :)

package-name-map/src/package-name-map.ts
justinfagnani: done

package-name-map/package.json
justinfagnani: changed to `1.0.0-pre.1` to be consistent with future prerelease bumps.

package-name-map/src/test/package-name-map_test.ts
justinfagnani: does it need to be the package-name-map's URL, or the document's baseURL?

domenic: Consider `https://example.com/a.html` with `<script type="packagemap" src="b/map.json">` which contains `pkg: { main: "c.js" }`.

Should `import "pkg"` give `https://example.com/b/c.js` or `https://example.com/c.js`? I'm not sure, but I lean toward the former; when I see `c.js` inside `https://example.com/b/map.json` I'd probably expect that to point to `https://example.com/b/c.js`.

package-name-map/src/package-name-map.ts
justinfagnani: Removed and redefined this function to perform successive URL resolution instead of path-joining.

package-name-map/src/package-name-map.ts
justinfagnani: done

package-name-map/src/test/package-name-map_test.ts
justinfagnani: done

package-name-map/src/test/package-name-map_test.ts
justinfagnani: done

package-name-map/src/test/package-name-map_test.ts
justinfagnani: I made the PackageNameMap constructor take a baseURI - it's agnostic as to where it comes from.

package-name-map/src/test/package-name-map_test.ts
justinfagnani: done

justinfagnani:
@domenic PTAL

cc @guybedford

package-name-map/src/package-name-map.ts
```diff
@@ -0,0 +1,278 @@
...
+    // 5. If no package entry is found, throw an error.
+    if (pkg === undefined) {
+      throw new Error(
+        `Unable to resolve specifier ${specifier} from referrer ${referrerURL}`
+      );
+    }
+
+    // 6. If the specifier is to a sub-module, but the package doesn't have a
+    //    path, throw an error.
+    if (specifier !== packageName && pkg.path === undefined) {
```
guybedford: I understand the sugars are still being fully decided, but if the default path is assumed to be the package name when setting a main, it seems a little inconsistent to then require an explicit path in the subpath case. Perhaps we should do both or neither instead of just doing the main assumption here?

package-name-map/src/package-name-map.ts
```diff
@@ -0,0 +1,278 @@
...
+ * decide whether validation errors should be early or late errors. Either way,
+ * it's still a TODO to implement a validate() method.
+ */
+export class PackageNameMap {
+  private _baseURI: string;
+  private _map: Scope;
+
+  constructor(map: Scope, baseURI: string) {
+    this._map = map;
+    this._baseURI = new URL('.', baseURI).href;
```
domenic: Use URL, not URI. See https://url.spec.whatwg.org/#url-apis-elsewhere

package-name-map/src/test/package-name-map_test.ts
```diff
@@ -0,0 +1,281 @@
...
+    suite('does not modify already valid specifiers', () => {
+      const map = new PackageNameMap(
+        {
+          packages: {
+            foo: {
+              main: 'index.js',
+            },
+          },
+        },
+        referrerURL
```
domenic: You should use a different base URL than the referrerURL. E.g. the base URL could be `https://example.com/resources/package-name-map.json` and the referrer URL could be `https://example.com/js/app.js`. That should help shake out some issues.

package-name-map/src/test/package-name-map_test.ts
```diff
@@ -0,0 +1,281 @@
...
+      );
+
+      test('does not modify URLs', () => {
+        assert.equal(
+          map.resolve('https://bar.com', referrerURL),
+          'https://bar.com'
+        );
```
domenic: This is not correct; the result should be the URL parsed-then-serialized. Which in this case gives `https://bar.com/` with a trailing slash.

package-name-map/src/test/package-name-map_test.ts
```diff
@@ -0,0 +1,281 @@
...
+            },
+            '@polymer/polymer': {
+              path: '@polymer/polymer',
+              main: 'polymer.js',
+            },
+          },
+        },
+        'http://foo.com/'
+      );
+
```
jdalton: How are percent encodings of package names handled? `"fs-thing"` vs. `"%66%73-thing"`?

domenic: Good question. Can you file a separate issue to make sure we add test coverage and consider it carefully? My intention is that package names should be treated as opaque and not decoded or encoded in that regard.

package-name-map/package.json
mkay581: Maybe publish it before someone takes the name? 😜

Issue #11, opened 2018-03-15
Field validations


guybedford:
What would be the exact form of the path, package name and scope name?

Enforcing all of these as _bare names_ might roughly do the job, but does it cover it sufficiently?

Further, if I have one invalid package name in "packages", or one invalid scope name in "scopes", does this validate on startup? How does it throw / warn?

Do all other validations happen during resolution itself? If so, do their validations throw as the resolution error?

domenic:
I don't really understand this issue. What is a bare name? What is an invalid package name? What is an invalid scope name?

The current validation is specified in https://github.com/domenic/package-name-maps#the-package-name-map-format.

guybedford:
I'm trying to think about what happens if I provide an absolute URL or full URL for a package name or scope name.

Bare names are as defined in the WhatWG loader spec resolver.

guybedford:
Also dot segments in package names or scope names could cause trouble here as well.

guybedford:
* And trailing `/` segments in scope or package names are valid or removed?
* And `\` in scope or package names? Do these get converted to `/` like URL resolution or do they invalidate?

domenic:
Good point on package names. I guess absolute URLs will probably never be hit, since the "resolve a module specifier" algorithm will bail out before reaching that package. I'm not sure it's worth adding a linting step that prohibits these. Unsure about ..s...

Scopes are not names, but URLs, so absolute URLs or ..s are fine and useful there.

My current plan is to write the algorithm (#6) and a bunch of main-stream test cases, then see what behavior falls out for weird cases like this, and then reconsider this issue.

guybedford:
> Scopes are not names, but URLs, so absolute URLs or ..s are fine and useful there.

Thanks for clarifying this, all examples made this look like a package name, but URL makes a lot of sense - running through the various use cases again this seems to work out pretty well.

Scopes would need to be resolved to full URLs when the manifest is loaded in order to do scope detection, so scopes that fail URL resolution would potentially result in errors on manifest load.

I only mentioned bare names as the closest thing we have to what validations a package name goes through currently.

Great, I'd be happy to run edge cases against the draft when you have something written up.

domenic:
The up-front validation looks quite nice. Although, it makes it clear you're not validating the paths yet at all.

[...]

PR #40, opened 2018-05-19
Test cases from #11 for reference implementation

guybedford:
I've tried to apply some of the cases mentioned in #11 to the reference implementation and managed to get some odd results for perusal.

Happy to discuss possible solutions here or in #11.

Great work @justinfagnani, the source is a pleasure to read and I was almost worried for a second there you'd managed to elegantly avoid all the edge case issues giving me nothing to complain about :)

### DIFF
--- a/package-name-map/src/package-name-map.ts
+++ b/package-name-map/src/package-name-map.ts
@@ -48,6 +48,56 @@ interface FindPackageResult {
 }
 
 /**
+ * Performs initialization-time scope validation
+ * Ensuring valid property types
+ */
+function validateScope(scope: Scope) {
+  if (scope.path_prefix !== undefined && typeof scope.path_prefix !== 'string') {
+    throw new Error(`path_prefix must be a valid string.`);
+  }
+  if (scope.packages !== undefined) {
+    if (typeof scope.packages !== 'object') {
+      throw new Error(`packages must be a valid object.`);
+    }
+    Object.entries(scope.packages).forEach(validatePackage);
+  }
+  if (scope.scopes !== undefined) {
+    if (typeof scope.scopes !== 'object') {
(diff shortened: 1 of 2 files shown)


guybedford:
(note that the test results being asserted here are the ones that make the tests pass, not what the expectations should necessarily be)

guybedford:
I've added a second commit here that adds some basic validations.

I guess the first question is as mentioned in #11 if validations should be performed early (during parsing) or at resolution time as resolution errors.

No stress at all if this implementation is entirely undesired - just trying to move conversations forward on this.

guybedford:
Sure, I've updated the basic type validations to happen on initialization of the package map. For now I've kept the package name validation to happen in this phase as well, until we can work out how to handle these cases.

I do think it would be better to carefully restrict package names to sensible cases by disallowing internal dot segments and trailing slashes, also in terms of users understanding the difference between package names and URLs. While yes these apply to the "package names as path" cases, they are still interpretations of the package name. Paths can be looser by definition so I do very much see these as package name validations.

For example, if a user were to mistakenly provide a URL-like into a package name (absolute URL, dot seegments). I've added the leading separator and URL case as well as a validation for now to demonstrate the usability here, which I think would be a nice cue for users to understand package names as something different.

If we do decide to go with more loose package name interpretations allowing these things without the stricter validations, that still leaves us with the trailing slash case which is problematic - `package-name/` as the package name:

* Should this trigger only for `package-name/` and `package-name/x` and not `package-name`?
* If so, `package-name/x` then has to be special-cased to ensure `x` appears in the output (https://github.com/domenic/package-name-maps/pull/40/commits/fb6db5c67cfda8114d2d85cc797cfb9399a6f446#diff-a3199dd5a8438364bee42f2d6ffc3826R379)

PR #43, opened 2018-06-17
Enable the default path for packages without a main

guybedford:
It seems a little inconsistent to apply the default path as the package name itself only when there is a main, when this approach can work just as well for submodules.

For example, this verbosity is already becoming clear in https://github.com/domenic/package-name-maps/compare/master...guybedford:package-name-path?expand=1#diff-a3199dd5a8438364bee42f2d6ffc3826R58.

This one thing will likely considerably simplify a lot of configurations in avoiding the need to use `path` entirely for most packages.

### DIFF
--- a/package-name-map/src/package-name-map.ts
+++ b/package-name-map/src/package-name-map.ts
@@ -102,18 +102,10 @@ export class PackageNameMap {
       );
     }
 
-    // 6. If the specifier is to a sub-module, but the package doesn't have a
-    //    path, throw an error.
-    if (specifier !== packageName && pkg.path === undefined) {
-      throw new Error(
-        `Cannot resolve specifier, no path found for package ${packageName}`
-      );
-    }
-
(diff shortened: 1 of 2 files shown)


domenic:
The body of this and the commit message seem at odds, right? This is actually enabling the default path for packages with *only* a main, from my read.

LGTM to merge with corrected commit message. (Or, if I'm misunderstanding, we should dig in a bit deeper...)

guybedford:
Yes, it's somewhat unclear - I've correct the commit message to reflect that this is about setting the path to the package name in the case that there is no path. So an empty object package would even work in such a case.

I do worry about empty objects as carrying meaning in the config though, so let me know if you want to discuss further on this.

domenic:
Looks good, thanks! I'm OK with empty objects carrying meaning; I'll work to add some tests to that effect.

Issue #3, opened 2018-03-15
Sugary defaults: picking a default value for "main"


domenic:
As noted [in one of the examples](https://github.com/domenic/package-name-maps#using-path_prefix-and-the-default-path), we could choose a default value for `"main"`, such as `${packageName}.js` or `index.mjs` or similar.

Doing so seems fraught, in that the browser would be picking preferred file names and extensions for your files. Apart from `favicon.ico`, and maybe a few other cases I'm forgetting, the browser has never done this before---for good reason, I think.

(What about index.html, you might say? Nope, the browser knows nothing about that convention. That's purely the server choosing to respond to `x/` with the on-disk-file `x/index.html`.)

But, I wanted to have a dedicated issue to discuss this.

matthewp:
I like the idea of allowing "main" to be optional, giving the server the freedom to define its own conventions like it does with index.html. It might do filesystem searching for index.js, index.mjs, packageName.js, etc. Or it might choose to serve a different main depending on browser capabilities.

bmeck:
If `"main"` is missing, is the intended behavior to resolve to a "directory":

```json
{
  "path_prefix": "/node_modules",
  "packages": {
    "moment": { },
  }
}
```

```mjs
import('moment'); // `/node_modules/moment/` ?
```

If so, the server could do proper redirecting for the exports (using `location:`, or `export *, {default}`).

bmeck:
@justinfagnani I'm not sure I understand how resolving to the directory would not allow them to do that.

justinfagnani:
Right, so I'm saying specifiers should not resolve to a directory.

bmeck:
I disagree since most servers let you do redirects already and find it simpler than trying to form a pattern relying on a specific path name that might change over time (`index.wasm` could be preferred pathname in the future!). Directories are not content-type specific, and already highly configurable.

justinfagnani:
Leaving it up to the server just means that other tools can't reliably use the mapping. I think it also puts more complexity on tools generating mappings, because they need to be aware of what the server might do for directories as well.

What about a specific opt-in to defaults per scope? Since the top level of a mapping is a scope, it's a natural place, and doesn't require any opinion by the platform.

bmeck:
@justinfagnani how would opt-in per scope be significantly different from just telling people to use `"main"`. Each package that differs since there isn't a default means this suffers the lack of reliability for package authors just like you describe for tooling authors.

justinfagnani:
[...]

Issue #4, opened 2018-03-15
Sugary defaults: using a string instead of just the package object


domenic:
[As noted in one of the examples](https://github.com/domenic/package-name-maps#using-path_prefix-and-the-default-path), we could allow shortening of `{ main: "moment.js" }` to just `"moment.js"`. Then you could write package map files like:

```json
{
  "path_prefix": "/node_modules/",
  "packages": {
    "moment": "moment.js",
    "html-to-text": "index.js",
    "redux": "lib/index.js"
  }
}
```

which is appealing for its simplicity. (Note how we still assume there's a path segment/folder corresponding to the package name, so e.g. redux is located at /node_modules/redux/lib/index.js. That follows from the default value for `"path"`.)

This slightly complicates the data model, as it would be the first instance of a union type in our structure. But it doesn't seem like that big of a deal. Should we do it?

justinfagnani:
This seems like a very nice ergonomics win for handwritten maps.

It's probably worth noting somewhere that package name maps are very similar to loader configs that have been around for a long time in loaders like RequireJS, etc. I think most of them have a simple path mapping like this.

guybedford:
I like this roughly in principle, but I would suggest not assuming the package name is in the path prefix. For example, packages can be installed under custom names in package.json files in npm.

Also consider users writing this sugar case by hand to point to CDN resources, and being confused when everything gets prefixed with the package name when they're providing a direct relative URL from the prefix base.

As mentioned in https://github.com/domenic/package-name-maps/issues/3#issuecomment-373470893 I'd prefer to be explicit, and also error when things are underspecified by distinguishing between file and directory cases, and treating this shorthand itself as a specific "file package" case and not a sugar.

domenic:
@guybedford we already are assuming the package name is the default path per https://github.com/domenic/package-name-maps#using-path_prefix-and-the-default-path . If you think we should remove that default (note: not a hard assumption, just a default) then perhaps open another issue.

I don't think we have any intention of supporting "file packages" in the current design. Remember, this is a package name map, not a module name map.

guybedford:
The path default is fine, but what I'm suggesting is that this sugar case could make it easy to handle a default case where users know where their modules are and just want to direct the packages to them.

For example if users have a few packages they want to load from a CDN and want to just write a map like:

```js
{
  "path_prefix": "https://cdn.com/'
  "packages": {
    "moment": "moment@2.3.4/index.js"
[...]

PR #44, opened 2018-06-17
Package main sugar

guybedford:
This adds the package main sugar case from #3, as well as testing some URL main cases which aren't currently tested.

Note if these types of feature PRs are too early, it's no problem to close and focus purely on the validation and test coverage diffs. This stuff is easy to add anytime.

### DIFF
--- a/package-name-map/src/package-name-map.ts
+++ b/package-name-map/src/package-name-map.ts
@@ -68,21 +68,28 @@ function validateScope(scope: Scope) {
     Object.values(scope.scopes).forEach(validateScope);
   }
 }
-function validatePackage([pkgName, pkg]: [string, Package]) {
+function validatePackage([pkgName, pkg]: [string, Package | string]) {
   // package name validation
   if (pkgName.startsWith('/') || pkgName.endsWith('/')) {
     throw new Error(
       `Invalid package name ${pkgName}, package names cannot start or end with a path separator.`
     );
   }
-  if (pkg.path !== undefined && typeof pkg.path !== 'string') {
-    throw new Error(
-      `Invalid package for ${pkgName}, path expected to be a string.`
-    );
+  if (typeof pkg === 'object') {
+    if (pkg.path !== undefined && typeof pkg.path !== 'string') {
+      throw new Error(
+        `Invalid package for ${pkgName}, path expected to be a string.`
+      );
+    }
+    if (pkg.main !== undefined && typeof pkg.main !== 'string') {
+      throw new Error(
+        `Invalid package for ${pkgName}, main expected to be a string.`
+      );
+    }
   }
-  if (pkg.main !== undefined && typeof pkg.main !== 'string') {
+  else if (typeof pkg !== 'string') {
     throw new Error(
-      `Invalid package for ${pkgName}, main expected to be a string.`
+      `Invalid package for ${pkgName}, must be a string or package object.`
     );
   }
 }
--- a/package-name-map/src/package-name-map.ts
+++ b/package-name-map/src/package-name-map.ts
@@ -16,7 +16,7 @@
  */
 export interface Scope {
   path_prefix?: string;
-  packages?: {[name: string]: Package};
+  packages?: {[name: string]: Package | string};
   scopes?: {[path: string]: Scope};
 }
 
@@ -259,7 +259,7 @@ const findPackage = (
     if (scope.packages) {
       for (const [pkgName, pkg] of Object.entries(scope.packages)) {
         if (isPathSegmentPrefix(pkgName, specifier)) {
-          foundPackage = pkg;
+          foundPackage = typeof pkg === 'string' ? {main: pkg} : pkg;
           foundPackageName = pkgName;
           break;
         }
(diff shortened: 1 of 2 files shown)


domenic:
This doesn't seem like it's implementing #3, but instead #4. We should be sure to note that correctly in the commit message.

I think the main thing we should also include here is what happens for submodules? Let us discuss that in #4 to keep things a bit focused. But I'd like to include tests for whatever we decide before merging this feature.

domenic:
I guess we already know what the submodule behavior should be; I got confused. But, I'd like us to add a few tests for it :)

domenic:
It seems like the type system is complaining now; not sure how to fix exactly.

We should also add tests for empty objects to this PR, I think.

guybedford:
Another question that comes up here, is should subpaths be supported for string mains?

```js
{
  packages: {
    pkg: '/path/to/main.js'
  }
}
```

In such a case we could:

* Set the "path" to dirname('/path/to/main.js')
* Set the "path" to undefined and support the subpath being "/pkg/[subpath]"
* Throw if trying to import a submodule for a package that is defined to be a string.

I'm tempted to say throwing seems the most well-defined, but it does involve treating packages as strings or objects deeper down into the codebase, unless we add a flag property to package "isMainSugar: true" or similar.

guybedford:
(I've got the typing fixes and test cases ready to go if we can clarify some of the above first)

domenic:
Hmm my take is we'd do the transformations in this way:

```
```js
{
  packages: {
    pkg: '/path/to/main.js'
  }
}
```

->

```js
{
  packages: {
    pkg: {
       main: '/path/to/main.js'
    }
  }
}
```

->

```js
{
  packages: {
    pkg: {
       path: 'pkg'
       main: '/path/to/main.js'
    }
  }
}
```

WDYT?

guybedford:
My worry here is the unintuitive split between URL mains and package names as paths.

For example, perhaps few would expect this to work when it would:

```js
{
  path_prefix: 'node_modules'
  packages: {
    pkg: 'index.js'
  }
}
```

where,

* `import 'pkg'` -> `node_modules/pkg/index.js`
* `import 'pkg/x'` -> `node_modules/pkg/x`

ljharb:
I’m not sure who would be using package name maps that would be entirely unfamiliar with node’s module system, or, who wouldn’t be able to rely on copious documentation explaining how name maps works.

domenic:
@guybedford Yeah, I see the problem. It seems like this is inherent in any solution that has implicit paths though. Or maybe it's just worse because of the string-valued version?

Maybe let's step back a bit.

```json
{
  "package1": {
    "path": "path1",
    "main": "main1.js"
  }
  "package2": {
    "main": "main2.js"
  },
  "package3": {
    "path": "path3"
  },
  "package4": { },
  "package5": "main5.js"
}
```

Then, according to our current plan, I think these are the results, right?

- `"package1"` -> `"https://example.com/path1/main1.js"`
- `"package1/sub"` -> `"https://example.com/path1/sub"`
- `"package2"` -> `"https://example.com/package2/main2.js"`
- `"package2/sub"` -> `"https://example.com/package2/sub"`
- `"package3"` -> PARSE ERROR missing main
- `"package3/sub"` -> PARSE ERROR missing main
- `"package4"` -> PARSE ERROR missing main
- `"package4/sub"` -> PARSE ERROR missing main
- `"package5"` -> `"https://example.com/package5/main5.js"`
- `"package5/sub"` -> `"https://example.com/package5/sub"`

Which of these seem like maybe we're making the wrong decision?

guybedford:
The main throwing errors are currently left out, I've added these in https://github.com/domenic/package-name-maps/pull/45.

To go over the cases:

```js
{
  "package1": {
    "path": "path1",
    "main": "main1.js"
  },
  "package2": {
    "main": "main2.js"
  },
  "package3": {
    "path": "path3"
  },
  "package4": { },
  "package5": "main5.js"
}
```

Then the corrected cases are:

1. `"package1"` -> `"https://example.com/path1/main1.js"`
2. `"package1/sub"` -> `"https://example.com/path1/sub"`
3. `"package2"` -> `"https://example.com/package2/main2.js"`
4. `"package2/sub"` -> `"https://example.com/package2/sub"`
5. `"package3"` -> PARSE ERROR missing main
6. `"package3/sub"` -> ~~PARSE ERROR missing main~~ `"https://example.com/package3/path3/sub"`
7. `"package4"` -> PARSE ERROR missing main
8. `"package4/sub"` -> ~~PARSE ERROR missing main~~ `"https://example.com/package4/sub"`
9. `"package5"` -> `"https://example.com/package5/main5.js"`
10. `"package5/sub"` -> `"https://example.com/package5/sub"` - this is currently undefined behaviour depending on how this PR works out

All of the cases do seem mostly sensible to me though, it's perhaps some of the main as a URL edge cases that get odd, but it's a super tough design space.

domenic:
OK, in that case I think we should either:

- Abandon this sugar for now as we're not sure about it yet, and we can always add it later; or
- Go with the original plan, i.e. 'Set the "path" to undefined and support the subpath being "/pkg/[subpath]"'.

guybedford:
I've updated to the original case here of setting `path` to undefined. I do think this main sugar is important to have for simple workflows and the edge case scenarios, which are growing on me, apply equally well to URL mains without a path in the non-sugar case anyway.

PR #45, opened 2018-06-27
Throw resolving a package with no main

guybedford:
This ensures we throw an error when resolving a package with a path but no main set, which wasn't previously tested.


domenic:
Hmm, I guess the main alternative here is failing at parsing time instead of resolution time. That is, say that a package is invalid if it doesn't have a main, and thus the whole map is invalid.

But, I can see how that might be too restrictive. So, let's do it this way.

PR #46, opened 2018-06-28
Update readme to reflect package main sugar (#4)

### LINKED ISSUE #4: Sugary defaults: using a string instead of just the package object

guybedford:
Closes #4.

### DIFF
--- a/README.md
+++ b/README.md
@@ -154,7 +154,18 @@ The above package name map can equivalently be written like so:
 
 _We've also considered defaulting `"main"` to `packagename.js` or `index.js`, but this would basically build a default file extension for JavaScript modules into the web, which is troublesome. Discuss in [#3](https://github.com/domenic/package-name-maps/issues/3)._
 
-_Another potential shortening is to allow e.g. `"src/moment.js"` as a shortcut for `{ "main": "src/moment.js" }`. The only downside here is that it complicates the data model by introducing a union type that needs to be normalized away. Discuss in [#4](https://github.com/domenic/package-name-maps/issues/4)._
+#### Package `"main"` sugar
+
+Providing a _string_ value directly for the package is treated as if setting just the main directly on the package:
+
+```json
+{
+  "path_prefix": "/node_modules",
+  "packages": {
+    "moment": "src/moment.js"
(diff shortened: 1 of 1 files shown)


README.md
```diff
@@ -217,20 +228,21 @@ Although we think using Web IDL is probably not a good idea for JSON formats, as
...
 }
 ```
 
 While parsing package name maps, we validate:
 - If non-strings show up where strings were expected, the map is invalid.
 - If non-objects show up where objects were expected, the map is invalid.
 - If the `"packages"` member is missing in a scope object, the map is invalid.
-- If the `"main"` member is missing in a package object, the map is invalid.
+- If any `"packages"` record keys start or end with a `/` separator, the map is invalid.
+- Any _string_ entries in the `"packages"` record are converted into package entries with a `"main"`.
```
domenic: Should we make this more comprehensively in tune with the reference implementation? Or maybe just link to the appropriate functions in the reference implementation so we don't have to keep them in sync while we're still iterating?

guybedford: This is actually something that is done differently between the implementation and the spec, in that the implementation doesn't actually clone the package map object and change it as we read it, while a C++ implementation would likely be constructing the package map data structure during the validation phase.

So I'd be weary of linking to the implementation on this one, and would rather refine the specification notes as the source of truth.

In the name of improving and making this more comprehensive, what else should be added do you think?

PR #50, opened 2018-08-18
Note package name map best practice

guybedford:
I don't think the HTTP/2 push comment is correct here as this would result in every HTML page visit getting the pushed package name map, whereas ideally this should be cached between page visits, or between page sections for multi-page apps.

### DIFF
--- a/README.md
+++ b/README.md
@@ -62,7 +62,7 @@ If you are surprised by this choice of solution, you may want to briefly visit t
 
 _We're not sure exactly how you install a package name map. The below represents a tentative idea. See the issue tracker for more discussion and alternatives: [#1](https://github.com/domenic/package-name-maps/issues/1)._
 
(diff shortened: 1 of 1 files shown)


domenic:
Apologies for letting this languish. I don't believe this advice is correct; in particular it implies a server round-trip for first-time visitors. Similar to the advice to inline critical CSS instead of putting it in external `<link>` elements, Chrome web performance folks believe that inlining your import map (or H/2 pushing it) makes more sense.

Hope that helps.

guybedford:
@domenic I think it is situation dependent - if its an SPA, then sure inline the import map, but if you are server-rendering your app and have a very large import map (multiple KB), then caching could be a more important factor.

domenic:
We've found that inlining critical resources is best practice regardless of application architecture; e.g. there are Lighthouse checks for doing so in the CSS case.

guybedford:
> We've found that inlining critical resources is best practice regardless of application architecture

If first-load is the optimization goal, then agreed that would be the result.

Issue #49, opened 2018-08-03
Should package name map resolution allow overriding URLs?


domenic:
In particular, given a package name map like the following:

```json
{
  "packages": {
    "https://example.com/foo": { "main": "node_modules/src/foo.js" },
  }
}
```

and a JS file like

```js
import "https://example.com/foo";
```

should the result import the file from `https://example.com/foo`, or from `node_modules/src/foo.js`?

Stated in spec terms, should the package name map resolution come before or after step 1 of [resolve a module specifier](https://html.spec.whatwg.org/multipage/webappapis.html#resolve-a-module-specifier)?

I really like the extra power of allowing overriding of URLs. It does depart from the literal sense of "package name map" though, and might imply a renaming of this project to something more like "module resolution map" or similar.

ljharb:
I’d expect a different top level key besides “packages” - maybe “urls” - but i also like the extra power.

domenic:
Hmm yeah. I guess it kind of ties into previous discussions around how much we want to restrict package names.

mkay581:
Hmm maybe we should wait until there is a strong use case for it? I can't seem to think of any.

domenic:
The strong use case is written up at https://github.com/drufball/layered-apis/issues/34 :). I'll write a new issue with a proposed solution.

PR #52, opened 2018-08-19
Provide a sensible default path for package main sugar case

guybedford:
The test case here is the following:

```json
  "path_prefix": "/node_modules",
  "packages": {
    "@polymer/polymer": "/dist/polymer/index.js",
  }
```

where we then test `resolve('@polymer/polymer')` as well as `require('@polymer/polymer/x')`.

The first case of the main resolution remains the same, but the second case is changed to provide a more sensible path default of the folder of the main used in the main sugar.

This way we get, with this new test:

```js
// Before the code change:
resolve('@polymer/polymer')    // -> "http://foo.com/dist/polymer/index.js"
resolve('@polymer/polymer/x') // -> "http://foo.com/node_modules/@polymer/polymer/x"

// After the code change:
resolve('@polymer/polymer')    // -> "http://foo.com/dist/polymer/index.js"
resolve('@polymer/polymer/x') // -> "http://foo.com/dist/polymer/x"
```

I've found in testing out the usage in practical scenarios that this helps to keep the package boundary well-defined and in line with intuition.

### DIFF
--- a/package-name-map/src/package-name-map.ts
+++ b/package-name-map/src/package-name-map.ts
@@ -266,7 +266,10 @@ const findPackage = (
     if (scope.packages) {
       for (const [pkgName, pkg] of Object.entries(scope.packages)) {
         if (isPathSegmentPrefix(pkgName, specifier)) {
-          foundPackage = typeof pkg === 'string' ? {main: pkg} : pkg;
+          foundPackage = typeof pkg === 'string' ? {
+            main: pkg,
+            path: pkg.substr(0, pkg.lastIndexOf('/')) || undefined
+          } : pkg;
           foundPackageName = pkgName;
           break;
         }
--- a/package-name-map/src/test/package-name-map_test.ts
+++ b/package-name-map/src/test/package-name-map_test.ts
@@ -144,7 +144,7 @@ suite('PackageNameMap', () => {
             moment: {
               main: 'http://moment.com/moment.js'
             },
-            '@polymer/polymer': 'index.js',
+            '@polymer/polymer': '/dist/polymer/index.js',
             '@polymer/polymer-foo': {
               main: '../polymer-foo.js'
             }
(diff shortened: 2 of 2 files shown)


package-name-map/src/package-name-map.ts
```diff
@@ -266,7 +266,10 @@ const findPackage = (
     if (scope.packages) {
       for (const [pkgName, pkg] of Object.entries(scope.packages)) {
         if (isPathSegmentPrefix(pkgName, specifier)) {
-          foundPackage = typeof pkg === 'string' ? {main: pkg} : pkg;
+          foundPackage = typeof pkg === 'string' ? {
+            main: pkg,
+            path: pkg.substr(0, pkg.lastIndexOf('/')) || undefined
```
ljharb: nit: it’d be better to use slice here - substr is in annex b

domenic:
The intention was that before the code change `resolve('@polymer/polymer/x')` would throw, hmm.

guybedford:
Ah, I can PR that if you like?

domenic:
Yeah, I dunno, still noodling over what the right thing is... also I got halfway through writing up a big issue to maybe solve #49 but then I realized my solution doesn't quite work... anyway part of my solution was to include your change in this PR, so I'm not sure. Let's keep this open for a bit to noodle on things?

guybedford:
I haven't read in depth but a couple of points that come to mind briefly on that:

1. `std:x` is maybe more like a bare specifier itself than a URL?
2. It can be possible to polyfill package name maps fully, down to changing the base_prefix to a supported module format (and this is exactly where I'm positioning SystemJS 2.0)

guybedford:
Another point on this - conditional loading was something I experimented with quite deeply, including various techniques from syntax `import 'a?b:c'` sort of thing to conditional declarations:

```js
map: {
  'x': {
    conditionA: 'xA/index.js',
    conditionB: 'xB/index.js'
  }
}
```

where conditions are loaded as modules themselves providing a boolean default export and get matched from first to last.

domenic:
OK, big issue almost written up; looking forward to your thoughts. One thing that becomes clear though writing examples in it, is that your choice here of basically making `"x/y/z"` sugar for `{ path: "x/y", main: "z" }` makes pretty much everything shorter and simpler. So I'm coming around.

domenic:
Sadly this PR has been outpaced by events. I hope we can resurrect the reference implementation soon!

Issue #64, opened 2018-11-03
Algorithm for handling chained and circular mappings


littledan:
How should a chained mapping like `{ "imports": { "/a": "/b", "/b": "/c" } }` behave? Does `import:/a` resolve to `/b` or `/c`? At least for merged maps, it could make sense for them to be composed and resolve to `/c`. But this could lead to some complexity.

What should happen in the case of a circularity, like `{ "imports": { "/a": "/b", "/b": "/a" } }` ? Maintain a list of already-visited things (sounds expensive)? Fail when resolving an import after 20 iterations, like redirects?

I see there's some logic to make the self-resolutions work, in the examples in the README, but I'm not sure exactly what that logic should be (just a one-off for that case?).

domenic:
The current intention is to keep things simple, and have the right hand side only allow non-`import:` URLs (designated as either absolute http(s) URLs, or by starting with `./`, `../`, or `/`), or built-in modules. `import:` URLs are not allowed on the right hand side, and so recursion shouldn't occur.

I _think_ this is consistent with the readme examples, but I admit things may have gotten lost as I shifted my thinking around.

littledan:
I see, thanks for explaining. I can believe this is consistent with the readme examples. If I think of a way to make this clearer in the README, I'll make a PR.

littledan:
To be clear, one consequence is that, if you have an imported module on the right-hand side, you're always referring the original one?

domenic:
I'm not sure what you mean by "imported module on the right-hand side", but if you have a URL there, you're referring to whatever is fetched from that URL.

littledan:
OK, this makes sense (I meant to say "built-in module", but either way, the meaning is clear).

PR #65, opened 2018-11-03
Clarify that mapping is never recursively applied

littledan:
Clarify the question from #64

### DIFF
--- a/README.md
+++ b/README.md
@@ -355,7 +355,7 @@ With this mapping, each class of browser behaves as desired, for our above impor
 - Browsers that support import maps, but do not support async local storage, will end up with a mapping from the polyfill URL to itself, and so will receive the polyfill anyway.
 - Browsers that support both import maps and async local storage will end up with a mapping from the polyfill URL to `@std/async-local-storage`, and so will receive the built-in module.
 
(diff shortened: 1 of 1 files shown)


littledan:
cc @guybedford , who was previously explaining to me how recursion of this map would be bad (I think it was unclear to both of us whether it was recursive).

README.md
```diff
@@ -355,7 +355,7 @@ With this mapping, each class of browser behaves as desired, for our above impor
...
 
-Note how we're using a capability here that we haven't explored in previous examples: remapping imports of "URL-like" specifiers, not just bare specifiers. But it works exactly the same. Previous examples changed the resolution of URLs like `import:lodash`, and thus changed the meaning of `import "lodash"`. Here we're changing the resolution of `import:/node_modules/als-polyfill/index.mjs`, and thus changing the meaning of `import "/node_modules/als-polyfill/index.mjs"`.
+Note how we're using a capability here that we haven't explored in previous examples: remapping imports of "URL-like" specifiers, not just bare specifiers. But it works exactly the same. Previous examples changed the resolution of URLs like `import:lodash`, and thus changed the meaning of `import "lodash"`. Here we're changing the resolution of `import:/node_modules/als-polyfill/index.mjs`, and thus changing the meaning of `import "/node_modules/als-polyfill/index.mjs"`. There is only one level of resolution for such cases; the target is no longer an `import:` URL, and therefore the import map is not consulted a second time.
```
ljharb: What happens if i put an import: url in the right hand side?

Issue #63, opened 2018-11-02
What exact characters are allowed in the map


domenic:
Both on the left and right-hand sides?

The only restriction stated in README.md right now is that the right-hand side either starts with `./`, `../`, or `/`, or is interpreted as a built-in module.

Do we want to add URL-based constraints? Does percent-decoding happen? Needs an algorithm, either for parsing or validation or both.

domenic:
The algorithm so far in https://github.com/domenic/import-maps/pull/67 normalizes and validates roughly so:

- No validation of the left-hand side
- Right hand side branches:
  - If it starts with `@std/`: `` new URL(`import:${rhs}`) ``
  - If it starts with `./`, `../`, or `/`: `` new URL(rhs, baseURL) ``
  - Try to parse as `new URL(rhs)` (NB: no base URL)
  - If parsing fails, or the parsed URL's scheme is not a fetch scheme, failure

I am happy with the RHS algorithm. For the LHS, probably we should not allow the empty string? But otherwise, as discussed in some previous issues, I think it should be pretty free-form... I should probably re-read them to make sure there wasn't something I missed though.

domenic:
Parser is merged, good to go here. The above remains accurate except now empty string specifier keys are disallowed.

PR #67, opened 2018-11-05
Start on a new reference implementation

domenic:
So far this includes the type-level schema validation and normalization logic. No domain-level validation is in place yet (i.e. checks on the contents of the strings involved; see #63).


domenic:
TODO: probably should normalize all right-hand sides into arrays (sometimes one-element arrays).

domenic:
The parser is much further along. It fully validates and normalizes the "specifier map" structure.

It it missing validation and normalization of the "scopes" structure, e.g. normalizing the keys as URLs.

domenic:
OK. Parser done, I am pretty sure. Going home for the day. Drive-by reviews appreciated, although given that this implementation is in service of the spec, understand that I might not take all advice on coding style or project setup.

I'll merge this in tomorrow or Thursday, and move on to the resolution algorithm. After that, time to translate into a more formal spec.

reference-implementation/lib/import-maps.js
```diff
@@ -0,0 +1,117 @@
...
+  }
+
+  if (string.startsWith('./') || string.startsWith('../') || string.startsWith('/')) {
+    return new URL(string, baseURL);
+  }
+
+  const url = tryURLParse(string);
+  if (url === null) {
+    return null;
+  }
```
guybedford: Out of interest, why can't both of the above branches just be represented by:

```js
try {
  return new URL(string, baseURL);
}
catch () {
  return null;
}
```

I thought absolute URL parsing in `new URL(absoluteURL, referrerURL)` was independent of the referrer?

reference-implementation/lib/import-maps.js
```diff
@@ -0,0 +1,117 @@
...
+
+  if (hasFetchScheme(url)) {
```
guybedford: I worry a little about this inhibiting experimentation with other protocols... what is the reason for this restriction?

reference-implementation/lib/import-maps.js
```diff
@@ -0,0 +1,117 @@
...
+};
+
+function normalizeSpecifierMap(obj, baseURL) {
+  delete obj[''];
+
+  for (const [key, value] of Object.entries(obj)) {
+    if (typeof value === 'string') {
+      obj[key] = [value];
+    } else if (!Array.isArray(value)) {
+      delete obj[key];
```
guybedford: Can we not throw on invalid right hand side types?

reference-implementation/lib/import-maps.js
```diff
@@ -0,0 +1,117 @@
...
+
+  // Always have these two keys, and exactly these two keys, in the result.
+  return {
+    imports: parsed.imports || {},
+    scopes: normalizedScopes
+  };
+};
+
+function normalizeSpecifierMap(obj, baseURL) {
+  delete obj[''];
```
guybedford: I would prefer to throw for empty string maps and scopes.

reference-implementation/lib/import-maps.js
```diff
@@ -0,0 +1,117 @@
...
+      if (!isJSONObject(specifierMap)) {
+        throw new TypeError(`The value for the "${scopeKey}" scope must be an object.`);
+      }
+
+      normalizeSpecifierMap(specifierMap, baseURL);
+
+      const scopeKeyURL = tryURLParse(scopeKey, baseURL);
+      if (scopeKeyURL === null) {
+        continue;
+      }
```
guybedford: I'd still prefer to be strict and throw errors for invalid cases.

reference-implementation/lib/import-maps.js
domenic: I find using try/catch for non-exceptional cases unfortunate, so I created this little wrapper. See https://github.com/whatwg/url/issues/372.

reference-implementation/lib/import-maps.js
domenic: Since it ignores them, I don't think it will inhibit experimentation. (See #80.) It's just that the initial spec will throw them away since they are unfetchable.

I don't think this will be observable except through side channels. E.g. you would probably see something in the dev console/network panel that fetches and fails if we kept them, whereas when we ignore them, you wouldn't see that. But since the fetch is always a failure anyway for non-fetch schemes, the result observable to the page would be the same.

reference-implementation/lib/import-maps.js
domenic: Empty string scopes seem useful. If you disagree please open a separate issue. For empty string specifier keys, and other cases where you'd prefer to throw, let's discuss in #80.

reference-implementation/lib/import-maps.js
guybedford: It was somewhat nitpicking the implementation details, but the suggestion was that:

```js
if (string.startsWith('./') || string.startsWith('../') || string.startsWith('/')) {
  return new URL(string, baseURL);
}
const url = tryURLParse(string);
if (url === null) {
  return null;
}
if (hasFetchScheme(url)) {
  return url;
}
return null;
```

could be replaced with the implementation-equivalent:

```js
const url = tryURLParse(string, baseURL);
if (url && hasFetchScheme(url)) {
  return url;
}
return null;
```

Although this is personal preference, and conceptual clarity should dominate too.

reference-implementation/lib/import-maps.js
guybedford: > Since it ignores them, I don't think it will inhibit experimentation. (See #80.) It's just that the initial spec will throw them away since they are unfetchable.

Right, but what happens if a browser implements a new protocol? Are fetch schemes extensible to support this?

guybedford: Say Node.js might want to implement its own internal schemes - if it were to implement package maps, it would then be "spec invalid"?

New protocols should not hit frictions in places like this.

The idea of locking down protocols just makes me weary in general, and I just don't understand what benefits we get for that.

domenic: Yes, I anticipate that Node.js will not implement the browser spec for import maps exactly, precisely because it takes a dependency on browser concepts like "fetch schemes".

Issue #61, opened 2018-10-29
If/then/else mappings for feature testing


littledan:
Continuing from [user story (F)](https://github.com/drufball/layered-apis/issues/34#issuecomment-427599908), it might be a bit suboptimal to have to use a top-level `await` to handle a case where you're testing for the existence of one module and want to take one of two paths depending on that (where it's not a simple fallback for the other module). Maybe we could provide a kind of if/then/else syntax to handle this case, which we could think of as the "desugared" form for the array pattern:

```
{
  "modules": {
    "globalize/list-format": {
      "if": "import:@std/intl-list-format",
      "then": "/node_modules/globalize/list-format-based-on-intl.mjs"
      "else": "/node_modules/globalize/list-format-from-scratch-with-data.mjs"
    }
  }
}
```

(We could decide this case isn't so important, but at TPAC, someone else raised the same issue as I had previously, and we began the discussion with Domenic explaining why depending on top-level await for fallback is not good for performance.)

littledan:
- I don't understand how `then` would conflict--the keys of that object would never be used for anything besides the if/then/else pattern. Anyway, the above was just a straw-person and we could do it in some other way.
- Let's stay away from jumping into JS here, for reasons explained in the README.

WebReflection:
* I guess my point is that resolving `then` module would look confusing with that proposal around
 * I don't think the README mention `if` conditional anywhere, and I wasn't advocating for JS anywhere but in there. The conditional import is otherwise somehow complicated when it comes to features detection due inability to have non-async imports. It'd be cool to solve that case, if possible.

littledan:
How should dynamic JS-generated maps create a conditional over which built-in modules are supported? By the time they are dynamically imported by a script, won't the realm be closed to adding import maps?

guybedford:
I think just supporting conditional fallback arrays already caters to this
use case ok.

littledan:
Fallbacks may work for Layered APIs that can be polyfilled, but if built-in modules are used for new web capabilities, you may want to perform feature detection to decide between different program paths. This was simple with globals, and it could be an unfortunate regression if you needed a jump through top-level await to get this fallback (when avoiding that top-level await is a goal of this proposal's fallback feature). Some hypothetical examples, if some existing browser APIs were done as built-in modules:
- Say IndexedDB were introduced as a built-in module. If it's not present in the browser, use a different codepath based on localStorage.
[...]

PR #82, opened 2018-12-04
Describe fallback based on top-level await

littledan:
Addresses #61

### DIFF
--- a/README.md
+++ b/README.md
@@ -19,7 +19,6 @@ _Or, how to control the behavior of JavaScript imports_
     - [For built-in modules, in module-import-map-supporting browsers](#for-built-in-modules-in-module-import-map-supporting-browsers)
     - [For built-in modules, in browsers without import maps](#for-built-in-modules-in-browsers-without-import-maps)
       - [This doesn't work for `<script>`](#this-doesnt-work-for-script)
-    - [Alternating logic based on the presence of a built-in module](#alternating-logic-based-on-the-presence-of-a-built-in-module)
   - [Scoping examples](#scoping-examples)
     - [Multiple versions of the same module](#multiple-versions-of-the-same-module)
     - [Scope inheritance](#scope-inheritance)
@@ -44,6 +43,7 @@ _Or, how to control the behavior of JavaScript imports_
   - [A convention-based flat mapping](#a-convention-based-flat-mapping)
 - [Adjacent concepts](#adjacent-concepts)
   - [Supplying out-of-band metadata for each module](#supplying-out-of-band-metadata-for-each-module)
+  - [Alternating logic based on the presence of a built-in module](#alternating-logic-based-on-the-presence-of-a-built-in-module)
 - [Implementation notes](#implementation-notes)
   - [`import:` URL staging](#import-url-staging)
   - [Further implementation staging](#further-implementation-staging)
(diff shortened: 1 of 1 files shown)


ljharb:
I’d be hesitant to include TLA in the readme - its current semantics are viral and may end up being untenable (in particular, they might make CJS support in ESM not possible, as well as having impacts in a pure ESM graph).

littledan:
@ljharb @jdalton Even if you don't like the solution, do you think the use case is realistic?

MylesBorins:
> in particular, they might make CJS support in ESM not possible

Unsure how this is the case when it is only supported in the module goal

ljharb:
My understanding is something like this: ESM A imports CJS B which requires ESM C, and C suddenly adds TLA, which would force B to be async, which would interfere with A's import of it.

MylesBorins:
@ljharb there is no current plan that would have ESM being able to be required in CJS... only option right now is dynamic import which is already async... I still fail to see this limitation

zenparsing:
There are practical module systems (e.g. `esm`) that allow CJS to depend directly on ESM source modules. Top-level await will present difficulties on such systems, widening the impedance mismatch between CJS and ESM rather than narrowing it.

Regarding this PR specifically, it highlights another concern that I have about TLA. By adding an attractively dynamic alternate import form (`await import(...)` ) we risk confusing users about which variant to use.

littledan:
Well, I proposed another solution in #61 which does not depend on top-level await. Let's discuss the problem there.

dcleao:
@ljharb, I think that @littledan was asking if you think that the described use case is _realistic_, and not if it is one of _your_ (node?) use cases...

I feel a constant push-back on TLA, or on any form of user-land JS being able to participate in module resolution, and possibly loading. As described in the [explainer](https://github.com/domenic/import-maps#a-programmable-resolution-hook), the presented reason is that switching between JS and native context is very expensive. That's surely objective, and I don't think anyone argues that that is true. However, that's not what's relevant, imo. Instead: how much slower is it compared to the alternatives that users will have to resort to to implement those use cases? Or, how worse is that compared to the impact that its absence causes on the shape of APIs?

We're flooded with valid, still current use cases. Just observe existing AMD/RequireJS applications (or SystemJS)! Definitely, these applications favor a JavaScript-centric approach to components.

Other proposed technologies, such as [HTML-modules](https://github.com/w3c/webcomponents/issues/645), [CSS-modules](https://github.com/w3c/webcomponents/issues/759) and [JSON-modules](https://github.com/w3c/webcomponents/issues/770) (let's create a spec for each file type!?) favor an HTML-centric approach to components.

I think that not having TLA or some form of loader plugin technology, it won't be possible to effectively develop JS-centric components (and probably also severely limit HTML-centric components).

Some use cases:

* **i18n loader plugin** — implement a way to ("statically") import a "JSON strings" resource file corresponding to the browser's current locale; the plugin can be configured with the list of locales for which resource files exist.

* **theme loader plugin** — load a different set of CSSs depending on the currently selected application theme; the plugin is configured with available theme names and corresponding CSS files/modules.

* **module configuration plugin** — load the configuration (similar to [out-of-band metadata](https://github.com/domenic/import-maps#supplying-out-of-band-metadata-for-each-module)) of a JS module; the effective configuration is the result of merging multiple JSON files/modules; the set of applicable files depends on the values of "global" variables of the current environment (which can change on each load; not defined at compile time (a fixed manifest, as suggested, doesn't work).

None of these would be possible to be implemented solely with client-side technologies (without dynamic, server-side endpoints).

Not everything is defined at static, compile time. The desired flexibility requires late-binding of configurations. And, we don't want that the asynchronous loading of configurations to leak asynchronicity (`Promise` in every return value?) to the APIs themselves.

CJS and the Node environment are not of the same nature of the Web environment. And the Web environment shouldn't become "jailed" by the nature of CJS.

I'd love to know the opinion of @guybedford on these subjects.

littledan:
This is an interesting discussion. Can we move it to the top-level await repo?

README.md
```diff
@@ -302,6 +303,28 @@ would not: in all classes of browsers, it would fetch the polyfill unconditional
 
 which will work as desired in all classes of browser.
 
+#### Alternating logic based on the presence of a built-in module
+
+_See further discussion of this case in the issue tracker: [#61](https://github.com/domenic/package-name-maps/issues/61)._
+
+Not all fallbacks take the role of running one piece of code. For example, sometimes, one code path is to be taken if a particular platform API exists, and another code path is taken if it doesn't exist. The import maps proposal does not aim to solve all such scenarios in a built-in way; instead, The Stage 2 TC39 proposal [top-level await](https://github.com/tc39/proposal-top-level-await) can be used to meet some of these use cases. Note that top-level await blocks import of the module containing the await.
```
domenic: It doesn't block import; it suspends execution like a normal `await` does. So I'll remove this last sentence as it seems to add confusion.

PR #86, opened 2018-12-11
doc: document that namespace are illustrative 

MylesBorins:
This proposal is currently using the `@${namespace}/` idiom, as currently
used by npm. We are in the process of discussing namespaces in Node.js core
and are debating between `@{namespace}/` and `${namespace}:`. This PR
changes the docs to use the url like schema which we can point to as
an example of cross ecosystem adoption of the pattern.

Refs: https://github.com/nodejs/node/pull/21551

### DIFF
--- a/README.md
+++ b/README.md
@@ -812,6 +812,11 @@ Several implementer questions come up around how `import:` URLs are envisioned t
 
 The high-level summary is that any fetch of an `import:` URL should be thought of as sugar for a series of if-then-else statements that in turn fetch the mapped-to URLs. For example, each fetch will pass through the service worker, until one succeeds.
 
+## Namespaces
+
+This proposal and the reference implementation, use the namespace `@std/` for illustrative purposes only. This proposal is generic and would
+be able to work with a variety of approaches to namespaces.
+
 ## Acknowledgments
 
 This document originated out of a day-long sprint involving [@domenic](https://github.com/domenic), [@hiroshige-g](https://github.com/hiroshige-g), [@justinfagnani](https://github.com/justinfagnani), [@MylesBorins](https://github.com/MylesBorins), and [@nyaxt](https://github.com/nyaxt). Since then, [@guybedford](https://github.com/guybedford) has been instrumental in prototyping and driving forward discussion on this proposal.
--- a/README.md
+++ b/README.md
@@ -242,6 +242,8 @@ import { StorageArea } from "@std/kv-storage";
 
 will first try to resolve to `@std/kv-storage`, i.e. the browser's built-in implementation of KV storage. If that fails, e.g. because the browser does not implement KV storage, then instead it will fetch the polyfill, at `/node_modules/als-polyfill/index.mjs`.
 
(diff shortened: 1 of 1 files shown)


ljharb:
We should really find a namespace name that doesn’t also mean “sexually transmitted disease”.

littledan:
Some history is noted in https://github.com/nodejs/node/pull/21551#issuecomment-446290404 . I think we'll need a bigger discussion to come to a strong conclusion here.

glen-84:
> We should really find a namespace name that doesn’t also mean “sexually transmitted disease”.

LOL, I had the same thought. 😄 

Probably need a bike-shedding issue for that.

* `system:date`
* `core:date`
* `js:date`
* `web:date`

domenic:
Thanks for this, but this repository is not the place to have this discussion; it's about a generic re-mapping mechanism. I'd appreciate it very much if people didn't let out their pent-up bikeshedding in this thread. We've talked in the past about starting a coordination repository for TC39 and the web to work together on the std library, and people should either wait for that to start up, or take this discussion to https://github.com/tc39/proposal-javascript-standard-library which will be one leg of those efforts and might be a good place to start the conversation while the web side of the work is still focused on making std modules polyfillable (i.e., the mechanisms in this repository).

I also find analogizing standards work to NSFW topics offensive and juvenile ("ha ha, sometimes you use the same letters as sex stuff when you write code, did you know that?"), and don't want to encourage that kind of behavior. This repository will shortly move to WICG where that will be against the [code of conduct](https://www.w3.org/Consortium/cepc/). Please conduct yourselves excellently.

As for what to do with the OP, I think the best thing is to introduce a comment to the README, shortly after the first usage of `@std/`, saying something about how this prefix is illustrative only, and not a part of the proposal. This is already in the reference implementation in https://github.com/domenic/import-maps/blob/0a7ba167f2050f0399ea325748b94d70049443dc/reference-implementation/lib/parser.js#L5 but you're right that upon re-reading it is not clearly stated in the README.

ljharb:
Apologies; I'm not intending to be juvenile, nor am i "giggling" about anything, but it seems I crossed a line anyways. I'll take further naming discussion to a different thread.

I think a clarification in the readme that it's a placeholder would be useful.

MylesBorins:
I've reverted the initial change and added a not to the readme

---

## Namespaces
 This proposal and the reference implementation, use the namespace `@std/` for illustrative purposes only. This proposal is generic and would
be able to work with a variety of approaches to namespaces.

glen-84:
> I also find analogizing standards work to NSFW topics offensive and juvenile ("ha ha, sometimes you use the same letters as sex stuff when you write code, did you know that?"), and don't want to encourage that kind of behavior.

I am the one that "laughed", and it was not about the abbreviation, but about the fact that Jordan and I had the same thought/concern.

While it's not the end of the world, I think that it's a fair comment, even if it wasn't made in the correct place. If you can avoid unfortunate naming, why not consider it while planning a new feature?

PR #111, opened 2019-03-08
Add caveats to extension-less imports example

domenic:
This is a follow-up to #109.

/cc @devsnek

### DIFF
--- a/README.md
+++ b/README.md
@@ -207,6 +207,12 @@ It is also common in the Node.js ecosystem to import files without including the
 
 would allow not only `import fp from "lodash/fp.js"`, but also allow `import fp from "loadsh/fp"`.
 
+Although this example shows how it is _possible_ to allow extension-less imports with import maps, it's not necessarily _desirable_. Doing so bloats the import map, and makes the package's interface less simple—both for humans and for tooling.
+
+This bloat is especially problematic if you need to allow extension-less imports within a package. In that case you will need an import map entry for every file in the package, not just the top-level entry points. For example, to allow `import "./fp"` from within the `/node_modules/lodash-es/lodash.js` file, you would need an import entry mapping `/node_modules/lodash-es/fp` to `/node_modules/lodash-es/fp.js`. Now imagine repeating this for every file referenced without an extension.
+
+As such, we recommend caution when employing patterns like this in your import maps, or writing modules. It will be simpler for the ecosystem if we don't rely on import maps to patch up file-extension related mismatches.
+
 ### Fallback examples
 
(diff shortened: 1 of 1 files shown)


README.md
```diff
@@ -207,6 +207,12 @@ It is also common in the Node.js ecosystem to import files without including the
 
 would allow not only `import fp from "lodash/fp.js"`, but also allow `import fp from "loadsh/fp"`.
 
+Although this example shows how it is _possible_ to allow extension-less imports with import maps, it's not necessarily _desirable_. Doing so bloats the import map, and makes the package's interface less simple—both for humans and for tooling.
+
+This bloat is especially problematic if you need to allow extension-less imports within a package. In that case you will need an import map entry for every file in the package, not just the top-level entry points. For example, to allow `import "./fp"` from within the `/node_modules/lodash-es/lodash.js` file, you would need an import entry mapping `/node_modules/lodash-es/fp` to `/node_modules/lodash-es/fp.js`. Now imagine repeating this for every file referenced without an extension.
```
ljharb: Since this is a common use case (most of npm atm), is it worth pursuing how important maps could meet it while avoiding the bloat (wildcards, namespaced suffixes, or something), instead of hoping that the wider community consistently reads and complies with this warning?

domenic: I think we have enough work to do with import maps that saving people from using extensions is not going to make it onto the use case list.

ljharb: I’m more suggesting that people are going to do this regardless, so if that incurs a size cost, it’d be useful for import maps to have a better story there. Understood that it’s not a priority.

jkrems: > For example, to allow `import "./fp"` from within the `/node_modules/lodash-es/lodash.js` file, you would need an import entry mapping `/node_modules/lodash-es/fp` to `/node_modules/lodash-es/fp.js`.

This currently reads like import maps apply after URL resolution. Is that correct? My understanding was that they apply *before* URL resolution..? Example:

* `"/foo/bar": "/foo/zapp"` - for the exact specifier "/foo/bar", replace with "/foo/zapp", then resolve URL based on base to `https://origin.com/foo/zapp`.

Or is the import map tried multiple times, once on the original string and once on the URL after resolution relative to the base? If it is only looked at before resolution relative to the base/referrer, then shouldn't `"./fp"` require its own unique entry in the import map and wouldn't be caught by `"/node_modules/lodash-es/fp"`?

domenic: If a specifier is URL-like, e.g. starts with ./ or ../, then URL resolution applies. See #112 or the reference implementation for more details.

PR #112, opened 2019-03-08
Add general URL-like specifier remapping examples

domenic:
As was noticed in #109, this was not an obvious capability up front.

---

Suggestions for more realistic examples welcome; I'm not super-happy with these.

### DIFF
--- a/README.md
+++ b/README.md
@@ -11,9 +11,10 @@ _Or, how to control the behavior of JavaScript imports_
   - [Bare specifiers](#bare-specifiers)
   - [Built-in modules](#built-in-modules)
 - [The import map](#the-import-map)
-  - [Bare specifier examples](#bare-specifier-examples)
+  - [Specifier remapping examples](#specifier-remapping-examples)
     - [Bare specifiers for JavaScript modules](#bare-specifiers-for-javascript-modules)
     - ["Packages" via trailing slashes](#packages-via-trailing-slashes)
+    - [General URL-like specifier remapping](#general-url-like-specifier-remapping)
     - [Extension-less imports](#extension-less-imports)
   - [Fallback examples](#fallback-examples)
     - [For user-supplied packages](#for-user-supplied-packages)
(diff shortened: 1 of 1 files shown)


README.md
```diff
@@ -191,6 +192,30 @@ import localeData from "moment/locale/zh-cn.js";
...
+
+```json
+{
+  "imports": {
+    "/app/helpers.mjs": "/app/helpers/index.mjs"
+  }
+}
+```
+
+This remapping ensures that any URL-like imports that resolve to `/app/helpers.mjs`, including e.g. an `import "./helpers.mjs"` from files inside `/app/`, or an `import "../helpers.mjs"` from files inside `/app/models`, will instead resolve to `/app/helpers/index.mjs`. This is probably not a good idea; instead of creating an indirection which obfuscates your code, you should instead just update your source files to import the correct files. But, it is a useful example for demonstrating the capabilities of import maps.
```
jkrems: > This remapping ensures that any URL-like imports that resolve to `/app/helpers.mjs`, including e.g. an `import "./helpers.mjs"` from files inside `/app/`, or an `import "../helpers.mjs"` from files inside `/app/models`, will instead resolve to `/app/helpers/index.mjs`.

Kind of repeating a comment in another thread - this suggests a "partial resolution" step that seems to be missing from the spec so far. It sounds like the idea is:

1. If `url` starts with `./` or `../`, set `url` to `new URL(url, referrer).path`.
2. Apply import mappings to `url`.
3. Assert: `url` isn't a bare specifier.
3. Return `new URL(url, referrer).href`.

Is that about correct?

domenic: More or less. See https://github.com/WICG/import-maps/tree/master/reference-implementation

jkrems: Ah, I missed the fact that the import map gets preprocessed in `parser.js` where `/foo/bar` keys get turned into fully resolved URL keys (if I'm scanning it correctly). Thanks for pointing me in the right direction. :)

domenic:
Going to merge, but would love more realistic examples if someone has ideas.

PR #113, opened 2019-03-11
Make most-specific specifier key win

### LINKED ISSUE #102: Matching with multiple trailing slash entries
```
{
  "imports": {
    "a": "/1",
    "a/": "/2/",
    "a/b": "/3",
    "a/b/": "/4/"
  }
}
```

`a/b/c` matches `a/` and `a/b/`, and thus might be mapped to `/4/c` or `/2/b/c`, respectively.
Also, `a/b` matches `a/` or `a/b`, and thus might be mapped to `/2/c` or `/3`, respectively.

Should we

- Reject this kind of import maps?
- Or refine the matching rule to remove ambiguity? For example, by using the longest match (in this case `a/b/c` should match with `a/b/`, and `a/b` should match with `a/b`).

The current ref impl seems to choose an matching entry based on the `Object.entries()` order.

domenic:
In particular, this matters in cases involving trailing-slash specifiers. Closes #102.

/cc @hiroshige as I implemented this notably different than https://chromium-review.googlesource.com/c/chromium/src/+/1475049.

### DIFF
--- a/reference-implementation/lib/parser.js
+++ b/reference-implementation/lib/parser.js
@@ -25,6 +25,12 @@ exports.parseFromString = (input, baseURL) => {
     normalizedImports = normalizeSpecifierMap(parsed.imports, baseURL);
   }
 
+  const sortedAndNormalizedImports = {};
+  const sortedImportsKeys = Object.keys(normalizedImports).sort(longerLengthThenCodeUnitOrder);
+  for (const key of sortedImportsKeys) {
+    sortedAndNormalizedImports[key] = normalizedImports[key];
+  }
+
   const normalizedScopes = {};
   if ('scopes' in parsed) {
     for (const [scopePrefix, specifierMap] of Object.entries(parsed.scopes)) {
@@ -49,7 +55,7 @@ exports.parseFromString = (input, baseURL) => {
 
   // Always have these two keys, and exactly these two keys, in the result.
   return {
-    imports: normalizedImports,
+    imports: sortedAndNormalizedImports,
     scopes: normalizedScopes
   };
 };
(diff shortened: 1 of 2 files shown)


reference-implementation/lib/parser.js
```diff
@@ -122,3 +128,17 @@ function normalizeAddress(address, baseURL) {
...
+
+function longerLengthThenCodeUnitOrder(a, b) {
+  return compare(b.length, a.length) || compare(a, b);
+}
+
+function compare(a, b) {
+  if (a > b) {
+    return 1;
+  }
+  if (b > 1) {
```
guybedford: Does this mean we don't sort `a` and `b` alphabetically in the import map, but we do sort `aa` and `bb` alphabetically?

reference-implementation/lib/parser.js
```diff
@@ -122,3 +128,17 @@ function normalizeAddress(address, baseURL) {
 function isJSONObject(value) {
   return typeof value === 'object' && value !== null && !Array.isArray(value);
 }
+
+function longerLengthThenCodeUnitOrder(a, b) {
+  return compare(b.length, a.length) || compare(a, b);
```
guybedford: Super edge case, but in the fallthrough case, `a` being `"##"` and `b` being `"00"` seems like it results in non-alphanumeric ordering since "00" > 1 :P

reference-implementation/lib/parser.js
domenic: This is a typo! Should be `b > a`. Now to write a test that exposes this...

reference-implementation/lib/parser.js
guybedford: Same issue as typo above then.

reference-implementation/lib/parser.js
guybedford: 👍

hiroshige-g:
> I implemented this notably different 

IIUC this difference doesn't cause any behavior differences, right?

domenic:
That's the intent, just thought it was interesting.

guybedford:
Also, the presorting implemented here is just a step away from the binary search algorithm optimization.

Issue #73, opened 2018-11-07
How do scopes match


domenic:
In particular substring matching is used by service worker and application manifest scopes, but this has been kind of weird and terrible, e.g. a scope of `https://google.com/maps` for that service worker also matches `https://google.com/mapsarecool`.

[Indeed, searching for "scope" in the app manifest issue tracker](https://github.com/w3c/manifest/issues?utf8=%E2%9C%93&q=is%3Aissue+is%3Aopen+sort%3Aupdated-desc+scope) reveals a lot of weird issues. This is a subtle area.

I think import maps are not quite as fraught, as they are not as cross-cutting: the scope concept here governs a specific type of string translation, and doesn't need to interact with larger web platform concepts like service workers or origins or so on. But we should tread carefully.

/cc @mgiuca. Probably I should schedule a brain-dump session with you to see what aspects of the manifest experience I can draw on.

domenic:
Related question came up while writing tests: if you use an empty string as your scope, what does that mean?

- Applies to all modules in the app, as if it were top-level `"imports"` instead of nested under `"scopes"`?
- Resolves the empty string against the base URL, i.e., use the base URL as the scope?
- Throw an exception and thus ignore the whole import map?
- Drop that particular scope, with a console warning?

Writing tests assuming URL resolution for now, but we should consciously decide an answer before closing this issue.

mgiuca:
Happy to sync up over VC.

One thing about manifest is that the decision wasn't made in isolation, but specifically for consistency with Service Worker scope matching. Not exactly "compatibility" (since the two scope matching algorithms never overlap), but just consistency so users won't have to think about two different matching rules. It's possible that Import Maps are far enough removed from SW/Manifest that the consistency argument vanishes.

On the SW side, IIRC the rationale for the basic string prefix match was literally performance. Not having to split on '/' but instead being able to do a simple prefix match. Seems like premature optimization to me, and probably doesn't apply to imports.

Relevant discussion: w3c/ServiceWorker#1272

marijnh:
In terms of performance, prefix match + a check of whether the match ends at a 'token boundary' (next to a slash or at the end of the string) should be plenty fast and less likely to trip people up

guybedford:
@marijnh what sort of matching algorithm do you envisage for that?

marijnh:
[...]

PR #115, opened 2019-03-17
Reference implementation of scopes

guybedford:
This provides a simple implementation of scoped resolution with the various fallback tests.

The gist of the approach is:
* Sort scopes, and scope imports just like imports
* Generalize the matcher into a `resolveImportsMatch` function which is used on scopes
* The first best matching scope is checked for import resolution, falling back to the global imports.

Because of the way scopes are normalized during parsing, we end up needing to check for `parentURL[normalizedScopeURL.length] === '/' || parntURL[normalizedScopeURL.length - 1] === '/'` because scopes may or may not end in '/'. One optimization here might be to normalize scopes to always be with or without the trailing slash (possibly without?), but that seemed to affect quite a few of the parsing tests when I tried it. I could certainly implement that as well though.

Any other thoughts re edge cases very welcome, but I think most are already covered by the base import maps themselves.

### DIFF
--- a/reference-implementation/__tests__/resolving-scopes.js
+++ b/reference-implementation/__tests__/resolving-scopes.js
@@ -11,21 +11,79 @@ function makeResolveUnderTest(mapString) {
 }
 
 describe('Mapped using scope instead of "imports"', () => {
-  const inTwoScopesURL = new URL('https://example.com/js/app.mjs');
-  const inOneScopeURL = new URL('https://example.com/app.mjs');
+  const jsNonDirURL = new URL('https://example.com/js');
+  const inJSDirURL = new URL('https://example.com/js/app.mjs');
+  const topLevelURL = new URL('https://example.com/app.mjs');
 
   it('should fail when the mapping is to an empty array', () => {
     const resolveUnderTest = makeResolveUnderTest(`{
       "scopes": {
-        "/js": {
+        "/js/": {
           "moment": null,
           "lodash": []
         }
       }
     }`);
 
-    expect(() => resolveUnderTest('moment')).toThrow(TypeError);
-    expect(() => resolveUnderTest('lodash')).toThrow(TypeError);
+    expect(() => resolveUnderTest('moment', inJSDirURL)).toThrow(TypeError);
+    expect(() => resolveUnderTest('lodash', inJSDirURL)).toThrow(TypeError);
+  });
+
+  describe('Exact vs. prefix based matching', () => {
+    it('should match correctly when both are in the map', () => {
+      const resolveUnderTest = makeResolveUnderTest(`{
+        "scopes": {
+          "/js": {
+            "moment": "/only-triggered-by-exact/moment",
+            "moment/": "/only-triggered-by-exact/moment/"
+          },
+          "/js/": {
+            "moment": "/triggered-by-any-subpath/moment",
+            "moment/": "/triggered-by-any-subpath/moment/"
+          }
+        }
+      }`);
+
+      expect(resolveUnderTest('moment', jsNonDirURL)).toMatchURL('https://example.com/only-triggered-by-exact/moment');
+      expect(resolveUnderTest('moment/foo', jsNonDirURL)).toMatchURL('https://example.com/only-triggered-by-exact/moment/foo');
+
+      expect(resolveUnderTest('moment', inJSDirURL)).toMatchURL('https://example.com/triggered-by-any-subpath/moment');
+      expect(resolveUnderTest('moment/foo', inJSDirURL)).toMatchURL('https://example.com/triggered-by-any-subpath/moment/foo');
+    });
+
+    it('should match correctly when only an exact match is in the map', () => {
+      const resolveUnderTest = makeResolveUnderTest(`{
+        "scopes": {
+          "/js": {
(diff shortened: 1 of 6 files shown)


reference-implementation/__tests__/resolving-scopes.js
```diff
@@ -0,0 +1,79 @@
...
+          "lodash-dot/": "./node_modules_2/lodash-es/",
+          "lodash-dotdot": "../node_modules_2/lodash-es/lodash.js",
+          "lodash-dotdot/": "../node_modules_2/lodash-es/"
+        }
+      }
+    }`);
+
+    const subScriptURL = new URL('https://example.com/app.mjs');
+
+    it('should resolve scoped and not cascade', () => {
```
domenic: What does "and not cascade" mean?

reference-implementation/__tests__/resolving-scopes.js
```diff
@@ -0,0 +1,79 @@
...
+        "/js": {
+          "lodash-dot": "./node_modules_2/lodash-es/lodash.js",
+          "lodash-dot/": "./node_modules_2/lodash-es/",
+          "lodash-dotdot": "../node_modules_2/lodash-es/lodash.js",
+          "lodash-dotdot/": "../node_modules_2/lodash-es/"
+        }
+      }
+    }`);
+
+    const subScriptURL = new URL('https://example.com/app.mjs');
```
domenic: Maybe `topLevelScriptURL` or `onlyInOneScopeURL`?

reference-implementation/__tests__/resolving-scopes.js
```diff
@@ -0,0 +1,79 @@
...
+
+    it('should resolve scoped and not cascade', () => {
+      expect(resolveUnderTest('lodash-dot')).toMatchURL('https://example.com/app/node_modules_2/lodash-es/lodash.js');
+      expect(resolveUnderTest('lodash-dotdot')).toMatchURL('https://example.com/node_modules_2/lodash-es/lodash.js');
+      expect(resolveUnderTest('lodash-dot/foo')).toMatchURL('https://example.com/app/node_modules_2/lodash-es/foo');
+      expect(resolveUnderTest('lodash-dotdot/foo')).toMatchURL('https://example.com/node_modules_2/lodash-es/foo');
+    });
```
domenic: What about `moment` with the default `scriptURL`? Does it use the `/` scope definition or not? I would think it should, but we need to test, either way.

reference-implementation/__tests__/resolving-scopes.js
```diff
@@ -0,0 +1,79 @@
...
+
+    it('should apply best scope match', () => {
+      expect(resolveUnderTest('moment', subScriptURL)).toMatchURL('https://example.com/node_modules_3/moment/src/moment.js');
+    });
+
+    it('should fallback to imports', () => {
+      expect(resolveUnderTest('moment/foo', subScriptURL)).toMatchURL('https://example.com/node_modules/moment/src/foo');
```
domenic: Good test.

reference-implementation/__tests__/resolving-scopes.js
```diff
@@ -0,0 +1,79 @@
...
+      expect(resolveUnderTest('lodash-dot', subScriptURL)).toMatchURL('https://example.com/app/node_modules/lodash-es/lodash.js');
+      expect(resolveUnderTest('lodash-dotdot', subScriptURL)).toMatchURL('https://example.com/node_modules/lodash-es/lodash.js');
+      expect(resolveUnderTest('lodash-dot/foo', subScriptURL)).toMatchURL('https://example.com/app/node_modules/lodash-es/foo');
+      expect(resolveUnderTest('lodash-dotdot/foo', subScriptURL)).toMatchURL('https://example.com/node_modules/lodash-es/foo');
+    });
+
+    it('should still fail for package modules that are not declared', () => {
```
domenic:
```suggestion
    it('should still fail for package-like specifiers that are not declared', () => {
```

reference-implementation/lib/parser.js
```diff
@@ -16,15 +16,9 @@ exports.parseFromString = (input, baseURL) => {
...
+  let sortedAndNormalizedImports = {};
   if ('imports' in parsed) {
-    normalizedImports = normalizeSpecifierMap(parsed.imports, baseURL);
-  }
-
-  const sortedAndNormalizedImports = {};
-  const sortedImportsKeys = Object.keys(normalizedImports).sort(longerLengthThenCodeUnitOrder);
-  for (const key of sortedImportsKeys) {
-    sortedAndNormalizedImports[key] = normalizedImports[key];
+    sortedAndNormalizedImports = normalizeSpecifierMap(parsed.imports, baseURL);
```
domenic: Maybe rename to sortAndNormalizeSpecifierMap

domenic:
I'm going to add to this PR to implement no-trailing-slash = exact matching, trailing-slash = prefix matching. For now that seems best to me. But let's keep #73 open and discuss more fully there; we can change after landing this.

reference-implementation/__tests__/resolving-scopes.js
guybedford: I see you worked this out! Seems good to have it too.

Issue #105, opened 2019-02-21
Import maps should be controlled by CSP


domenic:
I believe that, since the threat model of CSP is "someone is able to inject script into my page but I still want to prevent bad things from happening", we need import maps to be controlled by CSP as well. If a malicious actor can inject an import map, then they can change the behavior of scripts on the page (similar to overwriting `self.fetch()`, or inserting a `<base>` tag, or similar). So CSP should stop that as well.

This issue can be closed when there is an actual spec which includes these protections.

developit:
I think Import Maps should be subject to `script-src`, but should not be subject to `unsafe-inline`. An Import Map is not itself executable, it only allows redirecting script URLs to other URLs. Ultimately, those script URLs are external scripts and should be subject to any present `script-src` policy.

For example: say I have a CSP defined that blocks scripts not in a given list of sources:

```
Content-Security-Policy: unsafe-inline, script-src main.js libs.js
```

As long as the resulting mapped import is subject to CSP constraints, it's still possible to prevent loading of any unknown code via Import Maps:

```js
<script type="importmap">
{ "imports": {
  "@std/kv-storage": [
    "@std/kv-storage",
    "/kvs-polyfill.mjs"
  ]
} }
</script>
```

```js
import 'main.js'  // ok
import 'libs.js'   // ok
import '@std/kv-storage'  // violation
```

What's unclear to me, however, is whether the violation is triggered for `@std/kv-storage`, or `/kvs-polyfill.mjs`.  Intuitively, it seems like Built-in Modules should be unconditionally allowed and should not be affected by CSP, since they're part of the platform itself.

jeremyroman:
CSS also isn't executable, but IIUC `<style>` is still blocked if `unsafe-inline` is not in `style-src`.

Would it be so bad to just expect authors to use a hash or nonce in such cases (or to put import maps in a different CSP bucket, if that would be too cumbersome)?

hiroshige-g:
I feel we should apply at least all CSP restrictions that are currently applied to inline scripts also to import maps, because even though import maps are not "executable" in a strict meaning, they still can affect fetching and execution of other scripts.
(Not so sure this is sufficient though)

mikewest:
For script, CSP generally acts in two places:

1.  Inline script blocks are checked against nonces and `'unsafe-inline`'.
2.  Fetched resources are checked against the `script-src` safelist.

[...]

PR #119, opened 2019-03-22
Link mime-type notes to issue 105 (re CSP)

mikesamuel:
This incorporates the conclusion at the end of https://github.com/WICG/import-maps/issues/105#issuecomment-475330548

> We should probably not accept any JSON MIME type.

based on the assumptions made by existing hosting providers and JSON-producing web services as to what can be safely hosted on origins that appear on CSP source white lists.

### DIFF
--- a/spec.md
+++ b/spec.md
@@ -23,7 +23,7 @@ If a `<script type="importmap">` is encountered when _acquiring import maps_ is 
 Encountering a `<script type="importmap">` while _acquiring import maps_ is true will kick off a procedure roughly like this:
 
 1. If it's an external script (i.e. has a `src=""`), fetch it, using the usual "good defaults" of `<script type="module">`. (E.g., always UTF-8, "cors" mode, MIME type must match, ...)
-   - What should be the MIME type? Let's say `application/json+importmap`.
+   - What should be the MIME type? Let's say `application/importmap+json`.
      The required mime type should be distinct so that pre-existing services that assume `application/json`
      for uploaded content, and web services that might unintentionally produce something that looks like an
      importmap do not [enable CSP bypasses](https://github.com/WICG/import-maps/issues/105).
--- a/spec.md
+++ b/spec.md
@@ -23,7 +23,10 @@ If a `<script type="importmap">` is encountered when _acquiring import maps_ is 
 Encountering a `<script type="importmap">` while _acquiring import maps_ is true will kick off a procedure roughly like this:
 
 1. If it's an external script (i.e. has a `src=""`), fetch it, using the usual "good defaults" of `<script type="module">`. (E.g., always UTF-8, "cors" mode, MIME type must match, ...)
-  - What should be the MIME type? Let's say `application/json+importmap`? Maybe accept any JSON MIME type.
+   - What should be the MIME type? Let's say `application/json+importmap`.
+     The required mime type should be distinct so that pre-existing services that assume `application/json`
+     for uploaded content, and web services that might unintentionally produce something that looks like an
+     importmap do not [enable CSP bypasses](https://github.com/WICG/import-maps/issues/105).
 1. Parse the result as JSON into a spec-level struct (see below). (Will need to solve [whatwg/infra#159](https://github.com/whatwg/infra/issues/159) as part of this.)
 1. Merge the resulting struct into our realm's **merged import map** (see below).
 


spec.md
```diff
@@ -23,7 +23,10 @@ If a `<script type="importmap">` is encountered when _acquiring import maps_ is
 Encountering a `<script type="importmap">` while _acquiring import maps_ is true will kick off a procedure roughly like this:
 
 1. If it's an external script (i.e. has a `src=""`), fetch it, using the usual "good defaults" of `<script type="module">`. (E.g., always UTF-8, "cors" mode, MIME type must match, ...)
-  - What should be the MIME type? Let's say `application/json+importmap`? Maybe accept any JSON MIME type.
+   - What should be the MIME type? Let's say `application/json+importmap`.
```
Malvoz: It seems that the convention for MIME-type extensions would suggest `application/importmap+json` rather than `application/json+importmap` from skimming through https://www.iana.org/assignments/media-types/media-types.xhtml ^^

mikesamuel: @Malvoz, thanks for pointing that out.  I see many `...+xml` but no `...xml+...` that isn't also `...+xml`.  Fixed.

ljharb:
If ti has its own mime type, should it have its own extension as well? (I’m aware browsers wouldn’t pay attention to the extension)

mikesamuel:
@ljharb Having a separate extension would not help security-wise.

I'd be worried about this scenario:
1. a web service hosts uploaded files checks and uses [`mime.lookup`](https://www.npmjs.com/package/mime-types) to reject uploads of HTML, SVG, and JS files in accordance with [recommendations](https://www.owasp.org/index.php/Unrestricted_File_Upload).
1. the web service uses `mime.lookup`] to compute the Content-type header when serving an uploaded file
1. import-maps rolls out in browser with a `.importmap` extension
1. The maintainer updates to a new version of `mime.lookup` which maps `.importmap` to `application/importmap+json`.
1. An attacker uploads `payload.importmap` and exploits it via an injection vulnerability.

It's been best practice to not host third-party content on a sensitive origin for some time, but increasing the set of Content-type values that existing applications may unintentionally attach to hosted content is a risk factor.

ljharb:
@mikesamuel an extension is the typical way that servers decide which mime type to serve a file with; i'm not suggesting it would *help* security-wise, but i don't see how it'd hurt. (given that browsers would, just like with everything else, ignore the extension and only switch on mime type)

mikesamuel:
@ljharb 

> given that browsers would, just like with everything else, ignore the extension and only switch on mime type

Major frameworks fill in Content-type based on extension, so these are not orthogonal concerns.  For example, the widely used [send](https://www.npmjs.com/package/send#mime) package uses extension checking.  I believe that ends up being used to compute the mime-type for a lot of express apps, at least those that use serve-static.

> but i don't see how it'd hurt. 

I was trying to explain how in the scenario I just outlined.  Was that not clear?  Not convincing?

I don't think it's a big enough problem to warrant not recommending an extension.
But I also don't want to complicate this PR with something that will probably involve a lot of bikeshedding.

ljharb:
Fair that it doesn’t need to be in scope for the PR :-) but no, your example doesn’t seem any less problematic to me than if it uses the json extension.

mikesamuel:
@ljharb I don't think we disagree on any "... (should|must) ..." claims so unless you want me to expand on anything, have a great weekend.

mikesamuel:
> Fair that it doesn’t need to be in scope for the PR :-) but no, your example doesn’t seem any less problematic to me than if it uses the json extension.

@ljharb 
I think you're right.  AFAICT, my cases all assume things I've assumed are outside the threat model I outlined at https://github.com/WICG/import-maps/issues/105#issuecomment-475325493 : e.g. that there are few cases where an attacker can inject `<script type=importmap ...` but not `<script>...`.
Sorry for the confusion.

hiroshige-g:
https://github.com/WICG/import-maps/issues/105#issuecomment-475330548

> We should probably not accept any JSON MIME type.

@mikesamuel, Does this mean that "if a MIME type is accepted as a JSON, then it shouldn't be accepted as an import map (and vice versa)"?
If so, I noticed that `application/importmap+json` is a JSON MIME type and thus we should consider another MIME type.
https://mimesniff.spec.whatwg.org/#json-mime-type

(Or just meant that "not all JSON MIME types should be accepted as import maps"?)

PR #124, opened 2019-04-18
Fix and restructure the resolve algorithm to handle a missing case

domenic:
See individual commits for details.

### DIFF
--- a/reference-implementation/lib/resolver.js
+++ b/reference-implementation/lib/resolver.js
@@ -26,7 +26,7 @@ exports.resolve = (specifier, parsedImportMap, scriptURL) => {
     return asURL;
   }
 
-  throw new TypeError(`Unmapped bare specifier ${specifier}`);
+  throw new TypeError(`Unmapped bare specifier "${specifier}"`);
 };
 
 function resolveImportsMatch(normalizedSpecifier, importMap) {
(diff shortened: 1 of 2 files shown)


reference-implementation/lib/resolver.js
```diff
@@ -26,26 +26,33 @@ exports.resolve = (specifier, parsedImportMap, scriptURL) => {
...
-  throw new TypeError(`Unmapped bare specifier ${specifier}`);
+  throw new TypeError(`Unmapped bare specifier "${specifier}"`);
 };
 
 function resolveImportsMatch(normalizedSpecifier, importMap) {
   for (const [specifierKey, addressArray] of Object.entries(importMap)) {
-    if (addressArray.length > 1) {
-      throw new Error('Not yet implemented');
-    }
-    if (addressArray.length === 0 && specifierKey === normalizedSpecifier) {
```
guybedford: Is there a reason you couldn't just remove this second conditional? Although nicer to have good error messages too.

domenic: Well, it needs to be guarded, otherwise any empty arrays in the entire import map will cause any resolution to fail. This PR moves the guards outside, so we have the separate exact-match and prefix-match blocks.

guybedford: Ah right, so it only fails when it matches. Makes sense.

PR #125, opened 2019-04-18
Implement built-in module resolution and fallback

### LINKED ISSUE #120: Add reference implementation/tests for built-in module fallback
This isn't implemented in the reference implementation but it should be before it gets transcribed into a spec.

domenic:
Closes #120.

---

The test TODO in resolving-builtins.js is a bit troubling and I'd appreciate folks thoughts. We could leave it for now, but if people who haven't been staring at this code for hours have a good intuition, that'd be appreciated. /cc @guybedford @hiroshige-g.

### DIFF
--- a/reference-implementation/__tests__/resolving-builtins.js
+++ b/reference-implementation/__tests__/resolving-builtins.js
@@ -1,0 +1,118 @@
+'use strict';
+const { URL } = require('url');
+const { parseFromString } = require('../lib/parser.js');
+const { resolve } = require('../lib/resolver.js');
+const { BUILT_IN_MODULE_SCHEME } = require('../lib/utils.js');
+
+const mapBaseURL = new URL('https://example.com/app/index.html');
+const scriptURL = new URL('https://example.com/js/app.mjs');
+
+const BLANK = `${BUILT_IN_MODULE_SCHEME}:blank`;
+const NONE = `${BUILT_IN_MODULE_SCHEME}:none`;
+
+function makeResolveUnderTest(mapString) {
+  const map = parseFromString(mapString, mapBaseURL);
+  return specifier => resolve(specifier, map, scriptURL);
+}
+
+describe('Unmapped built-in module specifiers', () => {
+  const resolveUnderTest = makeResolveUnderTest(`{}`);
+
+  it(`should resolve "${BLANK}" to "${BLANK}"`, () => {
+    expect(resolveUnderTest(BLANK)).toMatchURL(BLANK);
+  });
+
+  it(`should error resolving "${NONE}"`, () => {
+    expect(() => resolveUnderTest(NONE)).toThrow(TypeError);
+  });
+});
+
+describe('Remapping built-in module specifiers', () => {
+  const resolveUnderTest = makeResolveUnderTest(`{
+    "imports": {
+      "${BLANK}": "./blank.mjs",
+      "${BLANK}/": "./blank/",
+      "${NONE}": "./none.mjs",
+      "${NONE}/": "./none/"
+    }
+  }`);
+
+  it('should remap individual built-in modules', () => {
+    expect(resolveUnderTest(BLANK)).toMatchURL('https://example.com/app/blank.mjs');
+    expect(resolveUnderTest(NONE)).toMatchURL('https://example.com/app/none.mjs');
+  });
+
+  it('should remap built-in module packages', () => {
+    expect(resolveUnderTest(`${BLANK}/foo`)).toMatchURL('https://example.com/app/blank/foo');
+    expect(resolveUnderTest(`${NONE}/foo`)).toMatchURL('https://example.com/app/none/foo');
+  });
+});
+
+describe('Remapping built-in module specifiers with fallbacks', () => {
+  const resolveUnderTest = makeResolveUnderTest(`{
+    "imports": {
+      "${BLANK}": ["${BLANK}", "./blank.mjs"],
+      "${BLANK}/": ["${BLANK}/", "./blank/"],
+      "${NONE}": ["${NONE}", "./none.mjs"],
+      "${NONE}/": ["${NONE}/", "./none/"]
+    }
+  }`);
+
+  it('should remap individual built-in modules', () => {
+    expect(resolveUnderTest(BLANK)).toMatchURL(BLANK);
+    expect(resolveUnderTest(NONE)).toMatchURL('https://example.com/app/none.mjs');
+  });
+
+  it('should remap built-in module packages', () => {
+    expect(resolveUnderTest(`${BLANK}/foo`)).toMatchURL(`${BLANK}/foo`);
+    expect(resolveUnderTest(`${NONE}/foo`)).toMatchURL('https://example.com/app/none/foo');
+  });
+});
+
+describe('Fallbacks with built-in module addresses', () => {
+  const resolveUnderTest = makeResolveUnderTest(`{
+    "imports": {
+      "blank": [
+        "${BLANK}",
+        "./blank-fallback.mjs"
+      ],
+      "blank/": [
+        "${BLANK}/",
+        "./blank-fallback/"
+      ],
+      "none": [
+        "${NONE}",
+        "./none-fallback.mjs"
+      ],
+      "none/": [
+        "${NONE}/",
+        "./none-fallback/"
+      ],
+      "blank-sub": [
+        "${BLANK}/sub",
+        "./blank-sub.mjs"
+      ]
+    }
+  }`);
+
+  it(`should resolve to "${BLANK}"`, () => {
+    expect(resolveUnderTest('blank')).toMatchURL(BLANK);
+  });
+
+  it(`should resolve to "${BLANK}" as a package`, () => {
+    // TODO is this correct? Seems incongruous with the blank-sub test case.
+    expect(resolveUnderTest('blank/foo')).toMatchURL(`${BLANK}/foo`);
+  });
+
+  it(`should fall back past "${NONE}"`, () => {
+    expect(resolveUnderTest('none')).toMatchURL('https://example.com/app/none-fallback.mjs');
+  });
+
+  it(`should fall back past "${NONE}" as a package`, () => {
+    expect(resolveUnderTest('none/foo')).toMatchURL('https://example.com/app/none-fallback/foo');
+  });
+
+  it(`should fall back past "${BLANK}" submodules`, () => {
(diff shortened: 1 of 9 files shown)


guybedford:
I must admit my intuition here was always that builtins wouldn't get subpathing support. Do you have any examples of builtins that would need this?

domenic:
Sure. For example, you may want a built-in element (say virtual scroller) which by default defines the element, when you do `import "std:virtual-scroller"`. But maybe you want to define it under a custom name; then you could do `import { HTMLVirtualScrollerElement } from "std:virtual-scroller/element"`, without the definition side-effects that the top-level module has.

We could just blanket prohibit built-in submodules for now, to keep things simple. (E.g. at parse time, throw away any std:-scheme URLs with slashes in them.) I think we may want them eventually, and I'd be a bit sad to not merge all the code and tests I wrote for that, but those aren't solid reasons :)

guybedford:
Perhaps restricting initially, then expanding when a well-defined resolution approach can be properly implemented makes sense? Concatenation would be inconsistent when it comes to backtracking here which doesn't seem ideal that eg:

```js
{
  "imports": {
    "virutal-scroller/": "std:virtual-scroller/"
  }
}
```

would allow `import('virtual-scroller/../another-builtin')` to resolve to `std:another-builtin` (if that is correct?).

That definitely would be bad, so I think in this case string concatenation could be a problem unless explicitly restricting these cases.

domenic:
Fortunately concatenation works as we'd hope in that case: https://jsdom.github.io/whatwg-url/#url=c3RkOnZpcnR1YWwtc2Nyb2xsZXIvLi4vYW5vdGhlci1idWlsdGlu&base=YWJvdXQ6Ymxhbms= , I guess because `std:` is not a [special scheme](https://url.spec.whatwg.org/#special-scheme).

guybedford:
Hmm so encodings aren't applied either? In that case, since the target domain is well-defined in the stdlib space, perhaps the bad cases are all well-defined failures then?

domenic:
Yeah, I think everything can be made to work fairly intuitively. The only case I'm unsure on right now is the one I pointed out in the TODO where I wasn't sure whether to apply fallbacks, if you map `"foo/": ["std:blank/", "/fallback/"]` but then resolve `"foo/bar"`. Should we return `"std:blank/bar"` (doesn't exist) or `"/fallback/bar/"`?

There may be other weird cases, and perhaps it'd be best to push solving them until later. That's the only one that stuck out to me immediately though.

guybedford:
That seems absolutely fine to me. You'd definitely want `/fallback/bar/` right, because the whole point of fallbacks is to handle stdlibs that don't exist now but could in future? While `std:blank/bar` might never exist, it seems best to have the fallback behave consistently for the whole scheme. Who's to say it won't be specified someday?

reference-implementation/__tests__/resolving-not-yet-implemented.js
```diff
@@ -0,0 +1,74 @@
...
+      ],
+      "bad3": [
+        "/bad3-1.mjs",
+        "/bad3-2.mjs"
+      ]
+    }
+  }`);
+
+  it('should fail for [built-in, built-in]', () => {
+    expect(() => resolveUnderTest('bad1')).toThrow(/Not yet implemented/);
```
hiroshige-g: "Not yet implemented" might be a little confusing, as it might mean "the support for this pattern of import maps is not yet implemented in the reference implementation" (which is what is intended here), or "the built-in module specified is not yet implemented". Adding a comment in this test might be helpful.

reference-implementation/__tests__/resolving-builtins.js
```diff
@@ -0,0 +1,118 @@
...
+    }
+  }`);
+
+  it(`should resolve to "${BLANK}"`, () => {
+    expect(resolveUnderTest('blank')).toMatchURL(BLANK);
+  });
+
+  it(`should resolve to "${BLANK}" as a package`, () => {
+    // TODO is this correct? Seems incongruous with the blank-sub test case.
+    expect(resolveUnderTest('blank/foo')).toMatchURL(`${BLANK}/foo`);
```
hiroshige-g: Do we allow this pattern?
This seems like we allow users to specify internal files under e.g. `std:kv-storage/foo-bar-impl.js`.
(Actually whether we allow import maps to map to `std:kv-storage/foo-bar-impl.js` and whether we allow actual access to `std:kv-storage/foo-bar-impl.js` is separate things, but probably it's better / easier to understand for users, to make them consistent.)

hiroshige-g:
> `import { HTMLVirtualScrollerElement } from "std:virtual-scroller/element"`

Does this mean we spec `std:virtual-scroller/element` in addition to `std:virtual-scroller`?
I also assumed that we don't allow any access or inspection into things behind `std:virtual-scroller`.

domenic:
Good points @hiroshige-g. Right, we would (hypothetically) spec `std:virtual-scroller/element`. We would *not* allow introspection into any of the (un-specced) implementation files. I can see how this is potentially more complexity.

Maybe we should just remove submodules of built-in modules for now and wait until we have a concrete spec that actually wants them. Both on the spec and implementation side, they will add complexity, and we should probably avoid paying that complexity until it's needed.

hiroshige-g:
+1 > remove submodules of built-in modules for now.

justinfagnani:
> Sure. For example, you may want a built-in element (say virtual scroller) which by default defines the element, when you do import "std:virtual-scroller". But maybe you want to define it under a custom name; then you could do import { HTMLVirtualScrollerElement } from "std:virtual-scroller/element", without the definition side-effects that the top-level module has.

@domenic to add support for removing submodules, if the motivation was a mainly the side-effect free import that didn't register the custom element, I think that would actually be a bad idea. Because of the 1-1 tagname-to-class restriction you don't want to allow someone to register a class that you might later.

In this case if module A imports `std:virtual-scroller/element` and defines is as `my-scroller`, and loads before module B which loads `std:virtual-scroller`, there would be an exception. Module A can still `import { HTMLVirtualScrollerElement } from "std:virtual-scroller"` and define a trivial subclass of it under a different name.

Even if/when we get scoped custom element registries, you still probably want to have the defining module reserve the name in the top-level scope.

hiroshige-g:
Just to provide more inputs, I'm wondering how import maps should behave on browsers with virtual-scroller versions that do not implement `std:virtual-scroller/element` yet.

If `std:virtual-scroller/element` triggers fallback based on whether `std:virtual-scroller` is implemented, then importing `std:virtual-scroller/element` via import maps fallbacks can still fail. This might require fallback mechanism outside import maps by e.g. dynamic imports.

If `std:virtual-scroller/element` triggers fallback based on whether `std:virtual-scroller/element` is implemented, then importing `std:virtual-scroller/element` (+ fallbacks) won't fail, but polyfilling might be a little more challenging.
For example, when user code imports both `std:virtual-scroller` and a polyfill of `std:virtual-scroller/element`, how can we polyfill? Just implementing everything in JavaScript in the polyfill is a simple solution, but this causes duplicated modules.

littledan:
I'm all for built-in element submodules (seems useful ergonomically), but I am interested to learn more about use cases for side effects on module load. We're sort of going back and forth in WebIDL-land about whether built-in modules should have a space to list side effects when they are loading. Where's the right place to discuss this?

domenic:
Hmm, well, the Web IDL PR makes sense to me. Personally I think side effects are a must, for the element-registration case alone.

domenic:
This is ready for another look, if folks have time!

reference-implementation/lib/parser.js
```diff
@@ -62,6 +62,10 @@ function sortAndNormalizeSpecifierMap(obj, baseURL) {
               `Package address targets must end with "/".`);
           return false;
         }
+        if (address.protocol === BUILT_IN_MODULE_PROTOCOL && address.href.includes('/')) {
+          console.warn(`Invalid target address "${address.href}". Built-in module URLs must not contain "/".`);
```
guybedford: I'm not sure about the outright banning of `/`, since it seems useful to be able to define names like `std:[spec]/[feature]`, which can be supported absolutely fine without running into subpath issues. It seems fine if this is just while these cases are being worked out, but I'm weary of this being enforced.

reference-implementation/lib/parser.js
```diff
@@ -113,6 +117,10 @@ function normalizeSpecifierKey(specifierKey, baseURL) {
 
   const url = tryURLLikeSpecifierParse(specifierKey, baseURL);
   if (url !== null) {
+    if (url.protocol === BUILT_IN_MODULE_PROTOCOL && url.href.includes('/')) {
+      console.warn(`Invalid specifier key "${url.href}". Built-in module URLs must not contain "/".`);
+      return null;
+    }
```
guybedford: Nit: perhaps abstract into a `validateBuiltinURL` function to avoid repetition of the error message.

reference-implementation/__tests__/resolving.js
```diff
@@ -169,12 +169,6 @@ describe('Mapped using the "imports" key only (no scopes)', () => {
       }
     }`);
 
-    it('should remap to built-in modules', () => {
-      expect(resolveUnderTest('/node_modules/als-polyfill/index.mjs')).toMatchURL('std:kv-storage');
-      expect(resolveUnderTest('https://example.com/node_modules/als-polyfill/index.mjs')).toMatchURL('std:kv-storage');
-      expect(resolveUnderTest('https://///example.com/node_modules/als-polyfill/index.mjs')).toMatchURL('std:kv-storage');
-    });
```
guybedford: Can we not keep this test, especially since it is implemented? Is this being removed because the resolver only knows about the `std:blank` builtin right now? Surely the list of supported builtins can be provided to the resolver initialization or as a host option?

reference-implementation/lib/parser.js
guybedford: Ahh, I see you've made the message specific to the case. NVM.

reference-implementation/__tests__/resolving.js
domenic: This test is fairly redundant with the newly-introduced ones, and yeah, it started failing once I made the implementation throw for anything not in `supportedBuiltInModules`. That will definitely be host-customizable, but for the reference implementation/tests, we don't need to make it so; the spec and tests will come out clear either way.

reference-implementation/lib/parser.js
domenic: I agree this would be nice. I'll file a follow-up issue, and maybe you or someone else can tackle fixing it :).

PR #135, opened 2019-06-02
Add general URL-like specifier remapping examples

Dewit62:
As was noticed in #109, this was not an obvious capability up front.


domenic:
This appears to be some kind of dupe of https://github.com/WICG/import-maps/pull/112

PR #140, opened 2019-06-21
Fix minor error handling issues around specifier resolution

hiroshige-g:
- Stop returning failure (instead of null) from
  `parse a URL-like import specifier`.
- Stop throwing inside `resolve a module specifier algorithm`, and
  return failure instead.
  Throwing an error causes e.g. throwing an error in WorkerGlobalScope
  that is not yet ready to execute scripts (
  https://html.spec.whatwg.org/C/#concept-environment-execution-ready-flag
  is false) when it occurred during `fetch a module worker script graph`.

***
<a href="https://pr-preview.s3.amazonaws.com/hiroshige-g/import-maps/pull/140.html" title="Last updated on Jun 25, 2019, 8:21 AM UTC (1ade17e)">Preview</a> | <a href="https://pr-preview.s3.amazonaws.com/WICG/import-maps/140/cbd5e7d...hiroshige-g:1ade17e.html" title="Last updated on Jun 25, 2019, 8:21 AM UTC (1ade17e)">Diff</a>

### DIFF
--- a/spec.bs
+++ b/spec.bs
@@ -176,7 +176,11 @@ At some point, each [=environment settings object=] will get an <dfn for="enviro
 <div algorithm>
   To <dfn lt="parse a URL-like import specifier|parsing a URL-like import specifier">parse a URL-like import specifier</dfn>, given a [=string=] |specifier| and a [=URL=] |baseURL|:
 
-  1. If |specifier| [=/starts with=] "`/`", "`./`", or "`../`", then return the result of [=URL parser|parsing=] |specifier| with |baseURL| as the base URL.
+  1. If |specifier| [=/starts with=] "`/`", "`./`", or "`../`", then:
+    1. Let |url| be the result of [=URL parser|parsing=] |specifier| with |baseURL| as the base URL.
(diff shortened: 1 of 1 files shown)


hiroshige-g:
@domenic PTAL when you have time.

hiroshige-g:
> Error-throwing

Er I see, so e.g.
https://html.spec.whatwg.org/multipage/webappapis.html#creating-a-javascript-module-script
should be modified so that it catches any errors during `resolve-a-module-specifier` and then set script's parse error to the caught error, right?
(For most of other callers we just have to turn "assert the result is not a failure" to "assert that no errors were thrown")
I think this is fine.

domenic:
Right. In particular, this makes it a bit clearer for implementers reading the spec what type of error information to propagate to dynamic import()s. Instead of the algorithm receiving "failure" and turning it into a generic TypeError, it receives a TypeError that (in the spec, at least) was generated where the failure occurred.

This is not an observable difference since error messages are not specified, but I think the flow is clearer.

PR #143, opened 2019-06-25
Make import maps registration priorities and error events in-order

### LINKED ISSUE #114: Order/Priority of multiple import maps
When there are multiple import maps present on a page, which one takes precedence? Is "document order" employed to determine the order?

Is the document order enforced at the moment when "acquiring import maps" is set to false? Or is document order enforced during initial loading of the page, and then dynamically injected import maps ignore document order and are applied in the order in which they are injected into the document?

## Scenario 1 - Two fetched import maps during initial load/processing
```html
<script type="importmap" src="/import-map1.json"></script>
<script type="importmap" src="/import-map2.json"></script>
<script type="module" src="import:widget"></script>
```

If import-map1.json takes longer to fetch than import-map2.json, which import map takes precedence? Meaning if both define the same module, which import map wins?

## Scenario 2 - No import maps during initial load, two dynamically injected in reverse document order

```html
<script>
  const importMap1 = document.createElement('script')
  importMap1.type = 'importmap'
  importMap1.textContent = JSON.stringify(getDynamicMap1FromSomewhere())
  document.head.appendChild(importMap1)

  const importMap2 = document.createElement('script')
  importMap2.type = 'importmap'
  importMap2.textContent = JSON.stringify(getDynamicMap1FromSomewhere())
  document.head.insertBefore(importMap2, importMap1)
</script>
<script type="module" src="import:widget"></script>
```

Import Map 2 is first in the DOM, but it was injected into the DOM second. Which one takes priority?

## Scenario 3 - Dynamic import injected after initial page load
```html
<script id="importmap1" type="importmap" src="import-map.json"></script>
<script>
  setTimeout(() => {
    const importMap = document.createElement('script')
    importMap.type = 'importmap'
    importMap.textContent = JSON.stringify(getDynamicMap1FromSomewhere())
    document.head.insertBefore(importMap, document.getElementById('importmap1'))
    import('import:widget')
  }, 3000)
</script>
```

Very similar to Scenario 2, except accentuating the difference in timing of when these import maps get put into the DOM. Which import map will take priority?

hiroshige-g:
Previously, import maps are processed when they are fetched, and therefore
the priority of import maps and order of `error` events are nondeterministic.

This PR introduces a mechanism to process import maps in the order of
`prepare-a-script`, fixes #114.

To do this, this PR
- Introduces `the script's import map`, `the import map is ready`,
  `list of pending import map scripts` and their related mechanisms similar
   to `the script's script`, `the script is ready`, and
   `list of scripts that will execute in order as soon as possible`.
- Replaces `pending import maps count` with
  `list of pending import map scripts`
- Moves fields (e.g. `acquiring import maps`) to `Document`
- Make `wait for import maps` work on `Document` directly instead of on
  `environment settings object`.

Other behavior changes:
- The check for errors in `parsing an import map string` now precedes the
  check for scripts elements moved between Documents.
- Import maps being loaded now delay the load event of Documents.

***
<a href="https://pr-preview.s3.amazonaws.com/hiroshige-g/import-maps/pull/143.html" title="Last updated on Jul 16, 2019, 6:35 PM UTC (36e1e65)">Preview</a> | <a href="https://pr-preview.s3.amazonaws.com/WICG/import-maps/143/988e9b3...hiroshige-g:36e1e65.html" title="Last updated on Jul 16, 2019, 6:35 PM UTC (36e1e65)">Diff</a>

### DIFF
--- a/spec.bs
+++ b/spec.bs
@@ -199,14 +199,14 @@ Inside the <a spec="html">prepare a script</a> algorithm, we make the following 
   <dl>
     <dt>"`importmap`"</dt>
     <dd>
-      [=Fetch an import map=] given <var ignore>url</var>, |base URL|, <var ignore>settings object</var>, and <var ignore>options</var>.
+      [=Fetch an import map=] given <var ignore>url</var>, |base URL|, |settings object|, and <var ignore>options</var>.
     </dd>
   </dl>
 - Insert the following case to <a spec="html">prepare a script</a> step 25.2:
   <dl>
     <dt>"`importmap`"</dt>
     <dd>
-      1. Let |import map parse result| be the result of [=parsing an import map string=], given <var ignore>source text</var> and |base URL|.
+      1. Let |import map parse result| be the result of [=create an import map parse result=], given <var ignore>source text</var>, |base URL| and |settings object|.
       1. Set [=the script's result=] to |import map parse result|.
       1. <a spec="html">The script is ready</a>.
     </dd>
@@ -248,7 +248,7 @@ Inside the <a spec="html">prepare a script</a> algorithm, we make the following 
     - The result of [=extracting a MIME type=] from |response|'s [=response/header list=] is not `"application/importmap+json"`
       <p class="note">For more context on MIME type checking, see <a href="https://github.com/WICG/import-maps/issues/105">#105</a> and <a href="https://github.com/WICG/import-maps/pull/119">#119</a>.</p>
   1. Let |source text| be the result of [=UTF-8 decoding=] response's [=response/body=].
-  1. Asynchronously complete this algorithm with the result of [=parsing an import map string=], given |source text| and |base URL|.
+  1. Asynchronously complete this algorithm with the result of [=create an import map parse result=], given |source text|, |base URL|, and |settings object|.
 
 </div>
 
@@ -334,9 +334,16 @@ To <dfn>register an import map</dfn> given an {{HTMLScriptElement}} |element|:
   1. If |parsed|["`scopes`"] [=map/exists=], then:
     1. If |parsed|["`scopes`"] is not a [=map=], then throw a {{TypeError}} indicating that the "`scopes`" top-level key must be a JSON object.
     1. Set |sortedAndNormalizedScopes| to the result of [=sorting and normalizing scopes=] given |parsed|["`scopes`"] and |baseURL|.
-  1. Let |import map| be a new [=/import map=] whose [=import map/imports=] are |sortedAndNormalizedImports| and whose [=import map/scopes=] scopes are |sortedAndNormalizedScopes|.
-  1. Return an [=import map parse result=] with [=import map parse result/settings object=] is ... [=import map parse result/import map=] is |import map|, and [=import map parse result/error to rethrow=] is ....
+  1. Return the [=/import map=] whose [=import map/imports=] are |sortedAndNormalizedImports| and whose [=import map/scopes=] scopes are |sortedAndNormalizedScopes|.
 </div>
+
+<div algorithm>
+  To <dfn>create an import map parse result</dfn>, given a [=string=] |input|, a [=URL=] |baseURL|, and an [=environment settings object=] |settings object|:
+
+  1. Let |import map| be the result of [=parse an import map string=] given |input| and |baseURL|. If this throws an exception, let |error to rethrow| be the exception. Otherwise, let |error to rethrow| be null.
+  1. Return an [=import map parse result=] with [=import map parse result/settings object=] is |settings object|, [=import map parse result/import map=] is |import map|, and [=import map parse result/error to rethrow=] is |error to rethrow|.
+</div>
+
 
 <div class="example" id="parsing-example">
   The [=/import map=] is a highly normalized structure. For example, given a base URL of `<https://example.com/base/page.html>`, the input
--- a/spec.bs
+++ b/spec.bs
@@ -183,7 +183,6 @@ Each {{Document}} has an <dfn for="Document">acquiring import maps</dfn> boolean
        1. If |result| is a failure, then set [=HTMLScriptElement/the script's import map=] to null.
        1. Otherwise, set [=HTMLScriptElement/the script's import map=] to the result of [=parsing an import map string=], given |result| and |base URL|.
        1. [=The import map is ready=].
-          <p class="note">If this makes all {{HTMLScriptElement}}s in [=Document/list of pending import map scripts=] [=HTMLScriptElement/the import map is ready|ready=], it will (asynchronously) unblock any [=wait for import maps=] algorithm instances.</p>
     1. Return.
 - Insert the following before <a spec="html">prepare a script</a> step 25.2:
   - If <a spec="html">the script's type</a> is "`importmap`", then:
@@ -202,6 +201,7 @@ Each {{Document}} has an <dfn for="Document">acquiring import maps</dfn> boolean
   1. If the element is not the first element in [=Document/list of pending import map scripts=] to which it was added above, then return without executing the script yet.
   1. <em>Register</em>: [=/Register an import map=] given the first element of [=Document/list of pending import map scripts=].
   1. Remove the first element of [=Document/list of pending import map scripts=].
+     <p class="note">If this makes the [=Document/list of pending import map scripts=] empty, it will (asynchronously) unblock any [=wait for import maps=] algorithm instances.</p>
   1. If the [=Document/list of pending import map scripts=] is still not empty and the first entry's [=HTMLScriptElement/the import map is ready=], then jump back to the step labeled <em>Register</em>.
 
 <p class="note">
--- a/spec.bs
+++ b/spec.bs
@@ -166,11 +166,11 @@ Each {{Document}} has an <dfn for="Document">acquiring import maps</dfn> boolean
 To process import maps in the <a spec="html">prepare a script</a> algorithm consistently with existing script types (i.e. classic or module), we make the following changes:
 
 - Introduce <dfn>import map parse result</dfn>, which is a [=struct=] with three [=struct/items=]:
-  - An <dfn for="import map parse result">settings object</dfn>, an [=environment settings object=],
-  - An <dfn for="import map parse result">import map</dfn>, an [=/import map=], and
-  - An <dfn for="import map parse result">error to rethrow</dfn>, a JavaScript value representing a parse error when non-null.
+  - a <dfn for="import map parse result">settings object</dfn>, an [=environment settings object=];
+  - an <dfn for="import map parse result">import map</dfn>, an [=/import map=]; and
+  - an <dfn for="import map parse result">error to rethrow</dfn>, a JavaScript value representing a parse error when non-null.
 - <a spec="html">the script's type</a> should be either "`classic`", "`module`", or "`importmap`".
-- Rename <a spec="html">the script's script</a> to <dfn>the script's result</dfn>, which can be either a <a spec="html">script</a> or an [=/import map=].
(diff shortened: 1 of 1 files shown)


hiroshige-g:
This PR is blocked by the discussion on #114 (not yet finalized).
This PR implements according to [2] of my comment
https://github.com/WICG/import-maps/issues/114#issuecomment-505133718.

joeldenning:
I like Option 2 because it doesn't rely on the "acquiring import maps" boolean. [I've expressed some other concerns](https://github.com/WICG/import-maps/issues/114#issuecomment-505429155) about Option 2, though, in the issue's comments. Overall I think Option 2 might be best.

hiroshige-g:
One issue in my mind is freedom of scheduling.

The current PR calls `register an import map` as early as possible.
For example, when an inline import map is inserted to DOM, the `error` event for that import map (on parse error) is required to be fired synchronously if there are no preceding pending external import maps.

However, probably `register an import map` is not so urgent things, and calling it anytime on or before `wait for import maps` is sufficient.

spec.bs
```diff
@@ -111,26 +117,44 @@ Each [=environment settings object=] has an <dfn for="environment settings objec
...
 <p class="note">For import maps, the script never becomes <a spec="html" lt="the script is ready">ready</a> and <a spec="html">the script's script</a> remains null.</p>
 
+<div algorithm="the import map is ready hook">
+  When [=HTMLScriptElement/the import map is ready=], run the following steps:
+
+  1. If the element is not the first element in [=Document/list of pending import map scripts=] to which it was added above, then return without executing the script yet.
+  1. <em>Register</em>: [=/Register an import map=] given the first element of [=Document/list of pending import map scripts=].
+  1. Remove the first element of [=Document/list of pending import map scripts=].
+     <p class="note">If this makes the [=Document/list of pending import map scripts=] empty, it will (asynchronously) unblock any [=wait for import maps=] algorithm instances.</p>
+  1. If the [=Document/list of pending import map scripts=] is still not empty and the first entry's [=HTMLScriptElement/the import map is ready=], then jump back to the step labeled <em>Register</em>.
```
domenic: Modern spec style is to use loops instead of gotos. Maybe tweak this to use a loop instead, if it's not too much work.

domenic:
> However, probably register an import map is not so urgent things, and calling it anytime on or before wait for import maps is sufficient.

I think it is kind of nice to get the error event sooner rather than later, but if you have a design that you think would be nicer I'm definitely open to it.

spec.bs
```diff
@@ -111,26 +117,44 @@ Each [=environment settings object=] has an <dfn for="environment settings objec
...
+
+<p class="note">
+  This is specified similarly to <a spec="html">list of scripts that will execute in order as soon as possible</a>, to register import maps and fire `error` events in order (<a spec="html">list of scripts that will execute in order as soon as possible</a> is almost never used in the wild though).
```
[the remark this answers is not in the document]
hiroshige-g: No. See Page 12 of https://docs.google.com/presentation/d/1Bvge7R6Yq6nzqjtf63sjmO714vBB3DmbhhIoDOPViVc/edit#slide=id.g337f25229d_0_181 for what kind of <script> elements fall within this category. (1.4% of total scripts, which is more than I expected when I wrote "almost never used" though)

non-async module scripts are processed by `list of scripts that will execute when the document has finished parsing`, which is another list of scripts.

spec.bs
hiroshige-g: Sounds totally fine. (I just copy-pasted the corresponding structure of the existing spec). I'll update.

hiroshige-g:
> Is it possible to reuse the script's script (maybe renamed "the script's result" or "the script element's result") and "the script is ready" (maybe renamed ("the script element is ready")? It seems like it would fit in fine since everything in step 26 is guarded by the type.
> If we were to follow the existing spec structure more closely, I think: ...

Sounds good, but needs more checks.

Probably, if we want to follow the existing spec structure, we can do one step further:

- Introduce a third subtype of `#concept-script`, i.e. `import map script`
- Make `fetch an import map` return an `import map script`
- `the script's script` is always a `script`
- Merge `register an import map` into `execute a script block`

This is what I and @nyaxt (IIUC) intentionally avoided, because the blocking semantics of import maps (i.e. ordering between import maps and other scripts) is governed quite differently from the existing mechanisms around `execute a script block`. For example, the order between classic/module scripts is controlled by the order of `execute a script block` calls, controlled by e.g. `list of scripts that will execute when the document has finished parsing`. On the other hand, the order between an import map and module script fetch are controlled by `wait for import maps` which is outside the mechanisms around `execute a script block`.
However, we've realized that we need a mechanism to control the order *between multiple import maps*, that controls `error` event order and priority of import maps. This can fit with the existing mechanisms around `execute a script block` (+ we'll still need the ordering mechanism for import-maps vs. scripts outside this), which might justify introducing the `import map script`. I missed this aspect mainly because we didn't consider multiple import map merging at early stages of package map design.

@domenic and @nyaxt, WDYT? (I'll anyway draft this tomorrow-ish, to see how it would look like)

domenic:
Hmm, interesting. I am OK either way, if you think we should go for that full type of merger, or just treat them more similarly inside prepare-a-script (e.g. by reusing`the script's script`).

spec.bs
hiroshige-g: Done.

hiroshige-g:
> > Is it possible to reuse the script's script (maybe renamed "the script's result" or "the script element's result") and "the script is ready" (maybe renamed ("the script element is ready")? It seems like it would fit in fine since everything in step 26 is guarded by the type.

> * Make `fetch an import map` return an `import map script`

Done.

> * Introduce a third subtype of `#concept-script`, i.e. `import map script`
> * `the script's script` is always a `script`
> * Merge `register an import map` into `execute a script block`

I didn't do these, because I found these don't simplify the spec further, and there are substantial differences between `#concept-script` and `import map script`.

domenic:
There is a problem with error handling. "Parse an import map" can throw an error. The current draft in this PR doesn't handle the error at all. I think it probably assumes that "parse an import map" would return null for parsing failures. But we shouldn't just convert any thrown errors into null as that loses the distinction between a parse error (window.onerror) and a fetch error (script.onerror).

I think the correct behavior would be to treat this similar to what we do for classic/module scripts, but simpler:

- Parsing errors should get stored as an "error to rethrow"
- "Prepare a script" should send such errors to window.onerror. (For classic scripts this is done in [run a classic script](https://html.spec.whatwg.org/multipage/webappapis.html#run-a-classic-script) 8.3.1. For our case the counterpart algorithm is "register an import map".)
- This probably means we need a structure that can hold an import map + possible error to rethrow, and that "the script's script" (which I just renamed "the script's result") should hold this structure. Naming this will be kind of annoying. One idea is "import map parse result" but maybe you have a better idea.

hiroshige-g:
Good catch. Then `import map parse result` would have:

- A settings object (I noticed the plumbing of settings object is also missing)
- An import map
- A parse error or an error to rethrow

This is more similar to a `script`, but still I'll keep this non-script.
(Blink implementation might have `ImportMapPendingScript` as a third subclass of `PendingScript`, but not `ImportMapScript` as a subclass of `Script`)

So far I don't have ideas for better names; I'll draft commits using `import map parse result` (and we can easily rename once we come up with a better name later).

domenic:
What will the settings object be used for?

hiroshige-g:
> What will the settings object be used for?

For checking whether the import map is moved between Documents.

hiroshige-g:
Hmm, firing `window.onerror` increases the motivation to make `import map parse result` a `script` and merge the caller of `register an import map` into `execute a script block` for consistency, and also makes me wonder about some issues (see `register an import map` in the latest commit).
I feel this is making `register an import map` too script-ish only just to fire error events...

(I'll think more about this later though)

spec.bs
```diff
@@ -268,13 +290,28 @@ Insert a call to [=wait for import maps=] at the beginning of the following HTML
...
    <p class="note">This is spec'ed consistently with <a href="https://github.com/whatwg/html/pull/2673">whatwg/html#2673</a>.</p>
    <p class="advisement">Currently we don't fire `error` events in this case. If we change the decision at <a href="https://github.com/whatwg/html/pull/2673">whatwg/html#2673</a> to fire `error` events, then we should change this step accordingly.</p>
-1. Otherwise, [=update an import map|update=] |element|'s [=node document=]'s [=Document/import map=] with |element|'s [=the script's result=].
+1. If |import map parse result|'s [=import map parse result/error to rethrow=] is not null, then:
+  1. <div class="issue">
+       Should we do the following, in order to match the behavior with <a spec="html">execute a script block</a>?
+       1. Increment/decrement <a spec="html">ignore-destructive-writes counter</a> of |element|'s [=node document=].
+       1. <a spec="html">Check if we can run script</a> with |settings object|. If this returns "`do not run`" then return.
+       1. <a spec="html">Prepare to run script</a> given |settings object|.
+       1. <a spec="html">Clean up after running script</a> given |settings object|.
```
domenic: I am pretty sure the only potentially observable thing here would be the "check if we can run script" step.

I guess that would be observable because it would make inserting import maps into non-fully active documents a no-op, which you could then observe by making the document fully active again? (Which I think can only happen with bfcache.) Wow, what an edge case.

spec.bs
```diff
@@ -268,13 +290,28 @@ Insert a call to [=wait for import maps=] at the beginning of the following HTML
...
+     </div>
+  1. <a spec="html">Report the exception</a> given |import map parse result|'s [=import map parse result/error to rethrow=].
+     <p class="issue">There are no relevant [=script=], because [=import map parse result=] isn't a [=script=]. Perhaps <a href="https://github.com/whatwg/html/issues/958">whatwg/html#958</a> might resolve this issue.</p>
```
domenic: Agreed, this is the best we can do until https://github.com/whatwg/html/issues/958 is solved.

spec.bs
```diff
@@ -297,7 +334,8 @@ To <dfn>register an import map</dfn> given an [=environment settings object=] |s
   1. If |parsed|["`scopes`"] [=map/exists=], then:
     1. If |parsed|["`scopes`"] is not a [=map=], then throw a {{TypeError}} indicating that the "`scopes`" top-level key must be a JSON object.
     1. Set |sortedAndNormalizedScopes| to the result of [=sorting and normalizing scopes=] given |parsed|["`scopes`"] and |baseURL|.
-  1. Return the [=/import map=] whose [=import map/imports=] are |sortedAndNormalizedImports| and whose [=import map/scopes=] scopes are |sortedAndNormalizedScopes|.
+  1. Let |import map| be a new [=/import map=] whose [=import map/imports=] are |sortedAndNormalizedImports| and whose [=import map/scopes=] scopes are |sortedAndNormalizedScopes|.
+  1. Return an [=import map parse result=] with [=import map parse result/settings object=] is ... [=import map parse result/import map=] is |import map|, and [=import map parse result/error to rethrow=] is ....
```
domenic: Probably this would work better as a wrapper algorithm that catches any errors thrown by "parse an import map"

domenic:
Yeah, I am torn. The "script" concept gets stretched pretty far, especially with new stuff like JSON module scripts.

If it helps, I made the following matrix:

| |Classic|JSON Module|JS Module|Import map|
|-|-------|-----------|---------|----------|
|Integrates with `<script>`|Y|*|Y|Y
|Inner data or parse error|Y|Y|Y|Y
|Fetch options/base URL for dependencies|Y| |Y| |
|Can be returned by fetch a single module script| |Y|Y| |

\* = I think this is currently allowed but there is talk of disallowing it? Can't find the issue

spec.bs
hiroshige-g: Introduced `create an import map parse result`.

spec.bs
domenic: I'll remove these for now.

PR #144, opened 2019-06-25
Update the README and spec.md for recent spec work

domenic:
Now that the actual specification (spec.bs) has become more solid, we can clean up some tentative notes, and delete most of spec.md.

Closes #119 by including the appropriate explanatory text in the README, instead of in spec.md.

### DIFF
--- a/README.md
+++ b/README.md
@@ -677,7 +677,7 @@ You can install an import map for your application using a `<script>` element, e
 <script type="importmap" src="import-map.importmap"></script>
 ```
 
-When the `src=""` attribute is used, the resulting HTTP response must have the MIME type `application/importmap+json`. (Why not reuse `application/json`? Doing so could [enable CSP bypasses](https://github.com/WICG/import-maps/issues/105).)
+When the `src=""` attribute is used, the resulting HTTP response must have the MIME type `application/importmap+json`. (Why not reuse `application/json`? Doing so could [enable CSP bypasses](https://github.com/WICG/import-maps/issues/105).) Like module scripts, the request is made with CORS enabled, and the response is always interpreted as UTF-8.
 
 Because they affect all imports, any import maps must be present and successfully fetched before any module resolution is done. This means that module graph fetching, or any fetching of `import:` URLs, is blocked on import map fetching.
 
--- a/README.md
+++ b/README.md
@@ -662,8 +662,6 @@ _Previous versions of this proposal anticipated making `import:` URLs resolve re
 
 ### Installation
 
-_The below represents a tentative idea. See the issue tracker for more discussion and alternatives: [#1](https://github.com/WICG/import-maps/issues/1)._
-
 You can install an import map for your application using a `<script>` element, either inline (for best performance) or with a `src=""` attribute (in which case you'd better be using HTTP/2 push to get that thing to us as soon as possible):
 
 ```html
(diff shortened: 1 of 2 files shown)


README.md
```diff
@@ -676,12 +674,14 @@ You can install an import map for your application using a `<script>` element, e
 ```
 
 ```html
-<script type="importmap" src="import-map.json"></script>
+<script type="importmap" src="import-map.importmap"></script>
```
guybedford: If this file extension is tentative, it could be worth including an additional note that the final file extension is still being decided. I hope it would be registered with IANA as well, as web servers would definitely need to support this. And I hope we can get that process going sooner rather than later too.

spec.md
```diff
@@ -1,35 +1,8 @@
-# Import maps proto-spec
+# Import maps stuff to formalize and merge into the spec
```
guybedford: Perhaps `semantics yet to be specified`?

I'd hope we can get some more references and thoughts to future features here too in due course.

README.md
domenic: File extension isn't intended to be specified; people can use whatever. Maybe I'll just omit it to be clearer.

guybedford: Hmm, that would be somewhat of an issue for adoption.

README.md
```diff
@@ -676,12 +674,14 @@ You can install an import map for your application using a `<script>` element, e
...
 ```
 
+External import maps must use the MIME type `application/importmap+json`. (Why not reuse `application/json`? Doing so could [enable CSP bypasses](https://github.com/WICG/import-maps/issues/105).)
```
joeldenning: "External import maps" means any import map downloaded from a different domain? Does this mean that "internal import maps" (from the same domain) do not need this MIME type?

I think that requiring that MIME type might be a huge barrier for adoption of import maps 😢 . Running a simple static server, or putting the file on a CDN, would no longer be good enough to serve an import map. Unless all popular static servers and CDNs are PR'ed to include a default mime type for `.importmap` files, which is enough work to push out adoption significantly.

domenic: It means import maps that are not inline in the document, but instead pulled from another URL.

I'm not significantly concerned about adoption of external import maps, as internal import maps will generally be the better pattern anyway.

guybedford: Tools that generate import maps will generate import maps into a file or stdout stream by default. Development workflows with import maps will naturally want to not have to edit the HTML template for simple operations like adding a new package, and instead use external import maps for ease of use. Production workflows can inline import maps where appropriate certainly, but disregarding entire development workflows has very real consequences for adoption.

joeldenning: > I'm not significantly concerned about adoption of external import maps, as inline import maps will generally be the better pattern anyway.

I disagree with this. For an ES module, it's nice to update the live version of your module without having to update an HTML file. Additionally, it's powerful for devs/organizations to "vend" a hosted version of their import map for consumption by other devs/organizations, which is something best done with an external import map imo.

joeldenning: At the organizations where I've helped them set up import maps, they host their js files on a CDN and want to be able to update the import map via CI process without having to redeploy the web server hosting the HTML file.

README.md
ljharb: If there’s a mime type, there should be an extension, whether browsers care about it or not - otherwise there’s no explicit reliable signal for servers to know which mime type to serve.

README.md
ljharb: Inlining any resource is generally considered a bad practice atm; I’m not sure why import maps would be any different.

domenic: Inlining resources is definitely not considered a bad practice. Like critical, above-the-fold CSS, import maps are best inlined.

I agree it can be easier for developers to not inline, given their workflows. The same is true for above-the-fold CSS. Regardless, it's better to inline.

README.md
domenic: I'm happy to leave that up to folks who work on servers to figure out.

README.md
ljharb: I'm pretty sure your position is not widely agreed upon, perhaps mine isn't either - at the very least, anything that can be inlined should be able to be externalized, and vice versa.

domenic: https://www.sitepoint.com/how-and-why-you-should-inline-your-critical-css/

ljharb: A lone article from 2.5 years ago doesn't prove or disprove either of our claims. This very website we're on right now doesn't have anything inlined, for example.

PR #150, opened 2019-07-10
Add spec for "update an import map"

domenic:
@hiroshige-g to review.

We'll need tests for this eventually, especially the sorting behavior and how that effects precedence, but I figured it'd be more straightforward to do those as web platform integration tests as part of the Chrome implementation, instead of doing them in the reference implementation.

***
<a href="https://pr-preview.s3.amazonaws.com/WICG/import-maps/pull/150.html" title="Last updated on Jul 11, 2019, 5:47 PM UTC (66b09cf)">Preview</a> | <a href="https://pr-preview.s3.amazonaws.com/WICG/import-maps/150/bd65e2d...66b09cf.html" title="Last updated on Jul 11, 2019, 5:47 PM UTC (66b09cf)">Diff</a>

### DIFF
--- a/spec.bs
+++ b/spec.bs
@@ -63,11 +63,67 @@ A <dfn>import map</dfn> is a [=struct=] with two [=struct/items=]:
 An <dfn>empty import map</dfn> is an [=/import map=] with its [=import map/imports=] and [=import map/scopes=] both being empty maps.
 
 <div algorithm>
-  To <dfn>update an import map</dfn> |import map| with a second [=/import map=] |new import map|:
+  To <dfn>update an import map</dfn> |importMap| with a second [=/import map=] |newImportMap|:
 
-  1. Assert: |import map| is not null.
-  1. Assert: |new import map| is not null.
-  1. <p class="note">TODO: Implement merging. This merges |new import map| into |import map| and thus updates |import map| in-place.</p>
+  1. For each |specifier| → |addresses| of |newImportMap|'s [=import map/imports=], [=map/set=] |importMap|[|specifier|] to |addresses|.
+  1. For each |url| → |specifierMap| of |newImportMap|'s [=import map/scopes=], [=map/set=] |importMap|[|url|] to |specifierMap|.
+  1. Set |importMap|'s [=import map/imports=] to the result of [=map/sorting=] |importMap|'s [=import map/imports=], with an entry |a| being less than an entry |b| if |a|'s [=map/key=] is [=longer or code unit less than=] |b|'s [=map/key=].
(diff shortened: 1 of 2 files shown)


spec.bs
```diff
@@ -63,11 +63,67 @@ A <dfn>import map</dfn> is a [=struct=] with two [=struct/items=]:
 An <dfn>empty import map</dfn> is an [=/import map=] with its [=import map/imports=] and [=import map/scopes=] both being empty maps.
 
 <div algorithm>
-  To <dfn>update an import map</dfn> |import map| with a second [=/import map=] |new import map|:
+  To <dfn>update an import map</dfn> |importMap| with a second [=/import map=] |newImportMap|:
 
-  1. Assert: |import map| is not null.
-  1. Assert: |new import map| is not null.
-  1. <p class="note">TODO: Implement merging. This merges |new import map| into |import map| and thus updates |import map| in-place.</p>
+  1. For each |specifier| → |addresses| of |newImportMap|'s [=import map/imports=], [=map/set=] |importMap|[|specifier|] to |addresses|.
```
hiroshige-g: |importMap|[|specifier|] and |importMap|[|url|] should be something like:

>  |importMap|'s scopes[|specifier|]
>  |importMap|'s imports[|url|]

spec.bs
```diff
@@ -63,11 +63,67 @@ A <dfn>import map</dfn> is a [=struct=] with two [=struct/items=]:
...
+  1. For each |url| → |specifierMap| of |newImportMap|'s [=import map/scopes=], [=map/set=] |importMap|[|url|] to |specifierMap|.
+  1. Set |importMap|'s [=import map/imports=] to the result of [=map/sorting=] |importMap|'s [=import map/imports=], with an entry |a| being less than an entry |b| if |a|'s [=map/key=] is [=longer or code unit less than=] |b|'s [=map/key=].
+  1. Set |importMap|'s [=import map/scopes=] to the result of [=map/sorting=] |importMap|'s [=import map/scopes=], with an entry |a| being less than an entry |b| if |a|'s [=map/key=] is [=longer or code unit less than=] |b|'s [=map/key=].
+</div>
+
+<div class="example" id="example-merging">
+  [=Update an import map=] merges the two import maps in a very simple way, not performing any deep merging beyond the top level of the "`imports`" and "`scopes`" keys. For example,
```
hiroshige-g: Although I feel a little inconsistent (because the `specifier map` as `import map`'s `imports` is merged while `specifier map`s inside `import map`'s `scopes` are overwritten, not merged), basically this behavior is fine with me.
If there are any discussions / analyses behind this decision, it will be helpful to reference them.

(In any cases, there might be still "unexpected" behavior around merging, because parts of different import maps can be applied to different parts of a single module script, but so far I have no idea for such kind of unexpected behavior, and probably we have to wait for user feedbacks.)

domenic: Not too much besides just a general drive for simplicity. I would be OK with slightly deeper merging. I will open an issue to discuss, but we should decide soon.

Issue #129, opened 2019-04-30
Reference implementation/spec: allow std:x/y built-in modules


domenic:
In #125, per some discussions there, I prevented any built-in modules that had slashes in their name. This was a bit over-ambitious. What we really want to prevent is built-in module package mappings, because that's pretty complicated.

This issue is tracking updating the reference implementation (and soon, spec) to be a bit more lenient in this regard. However, this is very non-urgent, as there are no built-in modules currently specified which would need this.

PR #152, opened 2019-07-11
Add access check of module URLs

hiroshige-g:
This commit reject the following URLs as request URLs of module scripts in all cases.

- Non-fetch scheme URLs.
- std: URLs with `/`.
- std: URLs of not implemented built-in modules.

These can't be the output of import maps since before
(i.e. the result of `resolve an imports match`),
but were allowed if directly specified in `<script>` or not mapped by import maps.

This commit also adds a non-normative placeholder for
referringScript-dependent access controls,
which is helpful for Chromium implementation.

This commit also adds support for the case where `referringScript`
is null in `resolve a module specifier`.

***
<a href="https://pr-preview.s3.amazonaws.com/hiroshige-g/import-maps/pull/152.html" title="Last updated on Jul 12, 2019, 6:30 PM UTC (f930690)">Preview</a> | <a href="https://pr-preview.s3.amazonaws.com/WICG/import-maps/152/988e9b3...hiroshige-g:f930690.html" title="Last updated on Jul 12, 2019, 6:30 PM UTC (f930690)">Diff</a>

### DIFF
--- a/spec.bs
+++ b/spec.bs
@@ -396,12 +396,16 @@ To <dfn>register an import map</dfn> given a [=string=] or a null |source text|,
   1. [=map/For each=] |scopePrefix| → |scopeImports| of |importMap|'s [=import map/scopes=],
     1. If |scopePrefix| is |baseURLString|, or if |scopePrefix| ends with U+002F (/) and |baseURLString| [=/starts with=] |scopePrefix|, then:
       1. Let |scopeImportsMatch| be the result of [=resolving an imports match=] given |normalizedSpecifier|, |scopeImports|, and |moduleMap|.
-      1. If |scopeImportsMatch| is not null, then return |scopeImportsMatch|.
+      1. If |scopeImportsMatch| is not null, then:
+        1. [=Check access=] given |scopeImportsMatch|, |settingsObject|, and |baseURL|.
+        1. Return |scopeImportsMatch|.
   1. Let |topLevelImportsMatch| be the reuslt of [=resolving an imports match=] given |normalizedSpecifier|, |importMap|'s [=import map/imports=], and |moduleMap|.
-  1. If |topLevelImportsMatch| is not null, then return |topLevelImportsMatch|.
+  1. If |topLevelImportsMatch| is not null, then:
+    1. [=/Check access=] given |topLevelImportsMatch|, |settingsObject|, and |baseURL|.
+    1. Return |topLevelImportsMatch|.
   1. <p class="note">At this point, the specifier was able to be turned in to a URL, but it wasn't remapped to anything by |importMap|.</p>
     If |asURL| is not null, then:
-    1. If |asURL|'s [=url/scheme=] is "`std`", and |moduleMap|[|asURL|] does not [=map/exist=], then throw a {{TypeError}} indicating that the requested built-in module is not implemented.
+    1. [=/Check access=] given |asURL|, |settingsObject|, and |baseURL|.
     1. Return |asURL|.
   1. Throw a {{TypeError}} indicating that |specifier| was a bare specifier, but was not remapped to anything by |importMap|.
 </div>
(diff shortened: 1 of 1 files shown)


hiroshige-g:
@domenic, PTAL when you have time.
Particularly, is this mechanism sufficient for getOriginals access restriction?

guybedford:
Does this mean that import maps cannot be used to map non-existent `std:[builtin]` URLs? That seems overly restrictive in a future where import maps are a baseline and this becomes a polyfilling technique as users would need to perpetually embed valid URLs to polyfill new builtin modules...

hiroshige-g:
> Does this mean that import maps cannot be used to map non-existent `std:[builtin]` URLs?

Mapping FROM non-existent `std:[builtin]` URLs: Can be used. We can use non-existent `std:[builtin]` URLs as import map keys (before and after this PR).

Mapping TO non-existent `std:[builtin]` URLs: Can't be used. In the existing spec https://wicg.github.io/import-maps/#resolve-an-imports-match there are no paths that can return non-existent `std:[builtin]` URLs.
Access check in this PR is applied to the OUTPUT of import maps, resolve a module specifier, not about INPUT or import map keys.
(Updated PR description)

domenic:
> This commit also adds support for the case where referringScript is null in resolve a module specifier.

This was a good catch, but I would like to land it separately and then put the new stuff here on top of it. I will post a PR shortly that extracts your work and also explains some other spec modifications needed.

domenic:
Overall this makes sense. I will rebase and rename "check access" to something like "validate module script URL".

Also note that we need to loosen the / restriction on built-ins (#129) especially now that there are [proposals like `std:elements/switch`](https://github.com/tkent-google/std-switch/). But for now we should keep the restriction.

Let me rebase it and stare a bit more, then I might merge or might have more comments.

domenic:
I spotted another preexisting problem that we should fix before inserting more code here. "resolve an imports match" throws an error but several of its callers assume that resolution failure results in a null return value.

I will work on another PR...

spec.bs
```diff
@@ -432,6 +436,23 @@ To <dfn>register an import map</dfn> given a [=string=] or a null |source text|,
   1. Return null.
 </div>
 
+<div algorithm>
+  To <dfn lt="validate the module script URL|validate a module script URL">validate a module script URL</dfn>, given a [=URL=] |url|, an [=environment settings object=] |settings object|, and a [=URL=] |base URL|:
+
+  1. <div class="note">Optionally, we can apply context-dependent access restrictions based on |base URL|. For example, Chromium might implement the followings, which aren't observable by users.
+         1. If |url|'s [=url/scheme=] is "`std-internal`" and |base URL|'s [=url/scheme=] is "`std-internal`", then return. This allows the use of "`std-internal`" within built-in modules, and rejects "`std-internal`" from user-provided scripts.
+         1. If |url|'s [=url/scheme=] is "`std`", |url| represents an access-restricted built-in module, and |base URL|'s [=url/scheme=] is not "`std-internal`", then throw a {{TypeError}}. This will be used to restrict access to getOriginals only within built-in modules.
```
domenic: The intention for our initial internal get-originals implementation is to use (1) here, not (2). I think (2) is a bit confusing and I would probably remove it.

spec.bs
```diff
@@ -396,12 +396,16 @@ To <dfn>register an import map</dfn> given a [=string=] or a null |source text|,
   1. [=map/For each=] |scopePrefix| → |scopeImports| of |importMap|'s [=import map/scopes=],
     1. If |scopePrefix| is |baseURLString|, or if |scopePrefix| ends with U+002F (/) and |baseURLString| [=/starts with=] |scopePrefix|, then:
       1. Let |scopeImportsMatch| be the result of [=resolving an imports match=] given |normalizedSpecifier|, |scopeImports|, and |moduleMap|.
-      1. If |scopeImportsMatch| is not null, then return |scopeImportsMatch|.
+      1. If |scopeImportsMatch| is not null, then:
+        1. [=Validate the module script URL=] given |scopeImportsMatch|, |settingsObject|, and |baseURL|.
```
domenic: I don't think this check will ever throw, since the import map will never return something that would fail the check.

spec.bs
```diff
@@ -396,12 +396,16 @@ To <dfn>register an import map</dfn> given a [=string=] or a null |source text|,
...
+        1. Return |scopeImportsMatch|.
   1. Let |topLevelImportsMatch| be the reuslt of [=resolving an imports match=] given |normalizedSpecifier|, |importMap|'s [=import map/imports=], and |moduleMap|.
-  1. If |topLevelImportsMatch| is not null, then return |topLevelImportsMatch|.
+  1. If |topLevelImportsMatch| is not null, then:
+    1. [=Validate the module script URL=] given |topLevelImportsMatch|, |settingsObject|, and |baseURL|.
```
domenic: Same here.

spec.bs
```diff
@@ -440,3 +461,17 @@ All call sites of HTML's existing <a spec="html">resolve a module specifier</a>
...
+
+<div algorithm="check-access-snippet">
+  1. [=Validate the module script URL=] given <var ignore>url</var>, |settings object|, and |settings object|'s [=environment settings object/API base URL=]. If this throws an error, then asynchronously complete this algorithm with null, and abort these steps.
+</div>
+
+at the beginning of the following HTML spec concepts (after [=wait for import maps=]):
+
+- [=fetch an external module script graph=]
+- [=fetch a modulepreload module script graph=]
+- [=fetch a module worker script graph=] (using <var ignore>module map settings object</var>)
```
domenic: It seems it might be cleaner to add the check at the beginning of "fetch a single module script". That way if we ever add new entry points (e.g. worklets?) those are covered automatically.

That would lead to redundant checks though, so engines probably wouldn't want to follow the spec exactly, which is annoying... I guess maybe the goal of easily adding new entry points is not going to work out. (E.g. they also need to wait for import maps.)

hiroshige-g:
Er, I was confused during rebasing. Still working on this and other PRs to fix commit chains.

hiroshige-g:
> Also note that we need to loosen the / restriction on built-ins (#129) especially now that there are proposals like std:elements/switch. But for now we should keep the restriction.

BTW I think the checks against URLs/specifiers should be centralized and simplified. Currently we have many kinds of URL/specifier checks in many places, which is hard to grasp for me. (I created a table:
https://docs.google.com/document/d/1m4wSpE0KyioTHtYK927RQoaANB_v1gRfyD1jitxmO7s/edit#heading=h.lf4qh3xpzry2 )

spec.bs
hiroshige-g: Sounds simpler.
So you mean get-originals have `std-internal:` URLs but not `std:` URLs, right?
In this way, this section just says "Chromium-internal things can't be observed from outside", which looks trivial.
(Anyway I'll leave this section, to guide implementation)

hiroshige-g: Removed (2).

domenic:
I agree with the drive for simplification. I'm unsure whether we should work to simplify before or after merging this patch.

spec.bs
hiroshige-g: Right, but for now I'd like to apply `validate the module script URL` on every paths.
Currently the checks that ensures this `validate the module script URL` never failing are scattered across the spec and it's not so clear what is rejected or not. I hope we can simplify/centralize the checks, but after this PR.

spec.bs
hiroshige-g: Yeah we can do this, if we allow redundant checks (i.e. calling validation twice for every non-toplevel fetch) as you mentioned. I'm a little concerned about the double checking pattern, because if a new non-const operation follows the same pattern in the future (e.g. it is added to the validation, or added to the spec in a similar fashion), it would be broken. (also I feel there was a similar bug due to doing something twice around dynamic import)

I switched to "fetch a single module script" approach, but still leave the original approach in a note.

hiroshige-g:
I think simplification should be done after this PR, to reduce ongoing PRs in parallel.

spec.bs
domenic: My motivation was to reduce the number of checks scattered throughout. But OK, in that case I'll add a comment to the source reminding us to remove this when we get a better grasp on things.

PR #164, opened 2019-07-22
Allow slashes in built-in module specifiers

### LINKED ISSUE #129: Reference implementation/spec: allow std:x/y built-in modules

domenic:
Closes #129.

***
<a href="https://pr-preview.s3.amazonaws.com/WICG/import-maps/pull/164.html" title="Last updated on Jul 30, 2019, 7:53 PM UTC (540f833)">Preview</a> | <a href="https://pr-preview.s3.amazonaws.com/WICG/import-maps/164/da5665c...540f833.html" title="Last updated on Jul 30, 2019, 7:53 PM UTC (540f833)">Diff</a>

### DIFF
--- a/reference-implementation/__tests__/parsing-addresses.js
+++ b/reference-implementation/__tests__/parsing-addresses.js
@@ -116,23 +116,19 @@ describe('Built-in module addresses', () => {
     );
   });
 
-  it('should ignore and warn on built-in module URLs that contain "/"', () => {
+  it('should allow built-in module URLs that contain "/" or "\\"', () => {
     expectSpecifierMap(
       `{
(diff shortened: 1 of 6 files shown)


hiroshige-g:
Step 1.1 of `validate a module script URL` should be also removed.
As for other parts LGTM.

domenic:
Good catch, thank you @hiroshige-g!

Issue #137, opened 2019-06-17
Cascading Import Map Resolution


michaelficarra:
## Problem

For the life of the web, JavaScript scripts have coordinated by mutating the global, either by adding properties that later-run scripts depend upon or by replacing properties that later scripts depend upon with ones that have the same API but augmented behaviour. We call the latter type of coordination "virtualisation".

As a concrete example of virtualisation, XMLHttpRequest is often virtualised by analytics tools, developer tools, performance monitors, [etc](https://github.com/domenic/get-originals/issues/18). When early-run scripts replace the reference to XMLHttpRequest on the global with a wrapper, two important things happen: 1) the later scripts do not have access to the original XMLHttpRequest constructor, so the early-run script can ensure they observe all usage, and 2) the augmented functionality composes with other virtualisers, in a last-to-first order.

Or consider polyfills: polyfilling scripts may augment built-in APIs with enriched forms. If multiple polyfills use the import maps API as specified today to virtualise a built-in, only the last-run polyfill will be applied. A more useful behaviour for import maps (and likely the one that would match developer intuition) would allow for the polyfill functionality to compose.

Finally, consider the [example](https://github.com/WICG/import-maps#denying-access-to-a-built-in-module) of denying access to a built-in module. It's natural to expect that a single import map early on the page which maps `"std:kv-storage"` to `null` would mean that access to "std:kv-storage" is denied to the rest of the page (just as when an early-run script deletes `window.localStorage` and so denies it to the rest of the page). But with the current strategy, that is not the case: a script later on the page can undo that mapping (i.e., add an import map mapping `"std:kv-storage"` to `"std:kv-storage"`) and then `import("std:kv-storage")` successfully.

## Proposal

Import maps, as specified today, default to later maps entirely overriding earlier ones when both provide mappings for the same module specifier, and thus do not support composition of behaviour. How could we change import maps to support these use cases? I propose lifting the restriction that entries in the right-hand side of an import map entry must be URLs, instead allowing any module specifier. Then, rather than destructively merging maps, the right side of each map would be resolved according to the previous map, ultimately falling out to the default resolution behavior. Call this cascading resolution.

This would mean that, if you have two maps along the lines of

```json
{
  "imports": {
    "foo": "bar"
  }
}
```

and

```json
{
  "imports": {
    "baz": "foo"
  }
}
```

on the page in that order, then the module specifier `"baz"` would resolve to `"bar"`, not `"foo"`.

[...]

Issue #153, opened 2019-07-11
Merging behavior: should we merge within scopes?


domenic:
(Previously the general topic was discussed in #88, but opening a new issue for this specific problem.)

The [current spec](https://wicg.github.io/import-maps/#update-an-import-map) defines merging very simply. Basically, you have

```js
const mergedImportMap = {
  imports: { ...importMap1.imports, ...importMap2.imports },
  scopes: { ...importMap1.scopes, ...importMap2.scopes }
};
```

In particular, this means that any scope definitions in the second import map completely override the first, as shown in the example linked in the spec.

In #150 @hiroshige-g pointed out that another potential design would be to always merge at the specifier map level. That would mean that if you have the same scope defined in two import maps, the contents of that scope get merged, instead of the second scope definition overriding the first.

I can see the attraction of such a design. I am OK switching, but would like community feedback.

My only worry is that I really do _not_ want to produce a more complex merging behavior. (For example, merging address arrays.) The current design is intentionally very simple and easy to model. I worry that by introducing something with a bit more complexity, we'll be making users expect that merging will be "smart" and "do what I mean", which is a target I do not want to aim for.

Thoughts? We should decide this soon.

For clarity let's discuss the alternatives as **top-level only merge** behavior or **merge-within-scopes** behavior.

guybedford:
As an additional point on merging - is ` { "scopes": { "x": null }` supported for nulling out previously defined scopes? I guess `"x": {}` can work just as well here though.

domenic:
You gotta use `{}`; any non-object will throw.

What that code will do, though, is an interesting question. In the top-level merge behavior, it will null out the previously-defined scope. In the merge-within-scopes behavior, it will be a no-op. (Similar to how you can't null out the top-level `"imports"` via a second import map with `{ "imports": {} }`.)

domenic:
Let me drop in a concrete example to try to make this issue more approachable. Given:

```html
<script type="importmap">
{
  "imports": {
    "a": "/a-1.mjs",
    "b": "/b-1.mjs",
    "std:kv-storage": ["std:kv-storage", "/kvs-1.mjs"]
  },
  "scopes": {
    "/scope1/": {
      "a": "/a-2.mjs"
    }
  }
}
</script>
<script type="importmap">
{
  "imports": {
    "b": null,
    "std:kv-storage": "kvs-2.mjs"
  },
  "scopes": {
    "/scope1/": {
      "b": "/b-2.mjs"
    }
  }
}
[...]

PR #167, opened 2019-07-24
add tests for import map composition and a "cascading" composition implementation

michaelficarra:
In the hopes that an implementation will help make the case for #137, I present an implementation of cascading import map composition and tests that demonstrate its behaviour.

***
<a href="https://pr-preview.s3.amazonaws.com/michaelficarra/import-maps/pull/167.html" title="Last updated on Mar 5, 2021, 9:39 PM UTC (22f5238)">Preview</a> | <a href="https://pr-preview.s3.amazonaws.com/WICG/import-maps/167/3467c3c...michaelficarra:22f5238.html" title="Last updated on Mar 5, 2021, 9:39 PM UTC (22f5238)">Diff</a>


domenic:
I've discussed this with some folks offline and we'd be willing to go this route if it doesn't add too much complexity.

However, to get a sense of that complexity, we'd need a more complete update to the reference implementation and tests. In particular, this only creates a merging procedure, but we also need to see and test how the resolution works, and how much complexity is added there. And we need to update the types and terminology used; right now the reference implementation represents addresses using `URL` objects, but if we're changing the right-hand side from a URL to a module specifier, then we need to stop using `URL` objects there.

In particular, in offline discussions you've explained how the mental model for web developers is something like

```js
// Turns relative URL-like specifiers into absolute-URL-like ones
let specifier = normalize(specifierFromSourceText);

specifier = apply(importMapB, specifier);
specifier = apply(importMapA, specifier);

// Throws on specifiers that are not absolute-URL-like. Identity transform otherwise.
let url = toURL(specifier);
```

We need to see this manifested in the reference implementation, where it will probably be something more like

```js
// Inside resolve() in resolver.js

// ... existing code to get normalizedSpecifier ...
const mappedSpecifier = apply(mergedImportMap, normalizedSpecifier, scriptURL);

const url = tryURLParse(mappedSpecifier);

if (!url) {
  throw new TypeError(`Unmapped bare specifier "${specifier}"`);
}

// ... existing code to reject un-implemented built-in modules ...

return url;
```

michaelficarra:
@domenic Take a look at the updated changes. I've updated/added tests and the resolving implementation.

reference-implementation/__tests__/composition.js
```diff
@@ -0,0 +1,367 @@
...
+  for (const mapLike of mapLikes) {
+    const newMap = parseFromString(JSON.stringify(mapLike), mapBaseURL);
+    map = appendMap(map, newMap);
+  }
+  return JSON.parse(JSON.stringify(map));
+}
+
+describe('Composition', () => {
+  it('should compose with the empty map on the left', () => {
+    const map = parseFromString(JSON.stringify({
```
domenic: I'd like to continue representing maps in the tests as strings.

reference-implementation/__tests__/composition.js
```diff
@@ -0,0 +1,367 @@
+'use strict';
+const { URL } = require('url');
+const { parseFromString } = require('../lib/parser.js');
+const { appendMap } = require('../lib/composer.js');
+
+const mapBaseURL = new URL('https://example.com/app/index.html');
+
+function composeMaps(mapLikes) {
```
domenic: It'd be nice to add a comment explaining that this functions serves a few purposes (as far as I can tell):

- Allowing the input to be plain JS objects, instead of import map objects or strings
- Automatically assumes the input will use mapBaseURL.
- Allowing writing tests that list all inputs, instead of creating the first, then appending xN

reference-implementation/__tests__/composition.js
```diff
@@ -0,0 +1,367 @@
...
+  if (!Array.isArray(mapLikes) || mapLikes.length < 1) {
+    throw new Error('composeMaps must be given a non-empty array of mapLikes');
+  }
+  let map = parseFromString(JSON.stringify(mapLikes.shift()), mapBaseURL);
+  for (const mapLike of mapLikes) {
+    const newMap = parseFromString(JSON.stringify(mapLike), mapBaseURL);
+    map = appendMap(map, newMap);
+  }
+  return JSON.parse(JSON.stringify(map));
```
domenic: What is the purpose of this? It seems like it's going to hide something from us...

reference-implementation/__tests__/composition.js
```diff
@@ -0,0 +1,367 @@
...
+        'https://a/': ['https://d/', 'https://g/', 'https://h/', 'https://f/'],
+        'https://b/': ['https://d/', 'https://g/', 'https://h/'],
+        'https://c/': ['https://f/'],
+        'https://e/': ['https://g/', 'https://h/']
+      },
+      scopes: {}
+    });
+  });
+
+  it('should compose maps that are using the virtualisation patterns we expect to see in the wild', () => {
```
domenic:
```suggestion
  it('should compose maps that are using the virtualization patterns we expect to see in the wild', () => {
```

reference-implementation/__tests__/composition.js
```diff
@@ -0,0 +1,367 @@
...
+      imports: {
+        'https://a/': ['https://b/'],
+        'https://c/': ['https://d/'],
+        'https://e/': ['https://f/']
+      },
+      scopes: {}
+    });
+  });
+
+  it('should compose maps that interact via cascading', () => {
```
domenic: Can you add some composition tests with relative URLs, specifically ones that are different representations of each other? For example, both /b and ../b refer to the same URL, and so if one map uses the former and another the latter, you should still get this kind of cascading behavior.

This should probably be done for the next several tests, at least until you feel confident we've covered the potential places that this could show up.

reference-implementation/__tests__/composition.js
```diff
@@ -0,0 +1,367 @@
...
+      imports: { 'std:built-in': ['https://built-in-enhancement-3/'] },
+      scopes: {
+        'https://built-in-enhancement-1/': { 'std:built-in': ['std:built-in'] },
+        'https://built-in-enhancement-2/': { 'std:built-in': ['https://built-in-enhancement-1/'] },
+        'https://built-in-enhancement-3/': { 'std:built-in': ['https://built-in-enhancement-2/'] }
+      }
+    });
+  });
+
+  it('should compose "nested" scopes', () => {
```
domenic: This seems to make a decision on https://github.com/WICG/import-maps/issues/153. Which I'm fine with; that issue is unresolved just because I've been waiting for someone to chime in, and you've done it. But can we add a simple test, like the one from (the scopes part of) https://github.com/WICG/import-maps/issues/153#issuecomment-513843388 , before getting in to cascading?

reference-implementation/__tests__/composition.js
```diff
@@ -0,0 +1,367 @@
...
+      {
+        imports: {
+          'moment/': '/node_modules/moment/src/'
+        },
+        scopes: {}
+      },
+      {
+        imports: {
+          'utils/': 'moment/lib/utils/',
+          'is-date': 'moment/lib/utils/is-date.js'
```
domenic: Can you test that this *doesn't* work within a single map? For example an entry

```js
'foo': 'utils/foo.js'
```

should result in... just a bare specifier in the output, I think?

reference-implementation/__tests__/composition.js
```diff
@@ -0,0 +1,367 @@
...
+        'moment/': ['https://example.com/node_modules/moment/src/'],
+        'utils/': ['https://example.com/node_modules/moment/src/lib/utils/'],
+        'is-date': ['https://example.com/node_modules/moment/src/lib/utils/is-date.js'],
+        'is-number': ['https://example.com/node_modules/moment/src/lib/utils/is-number.js']
+      },
+      scopes: {}
+    });
+  });
+
+  it('should url normalize things which have composed into URLs', () => {
```
domenic:
```suggestion
  it('should URL-normalize things which have composed into URLs', () => {
```

reference-implementation/__tests__/parsing-addresses.js
```diff
@@ -49,14 +49,14 @@ describe('Relative URL-like addresses', () => {
...
-        slash: [expect.toMatchURL('https://base.example/')]
+        dotSlash: ['https://base.example/path1/path2/'],
+        dotDotSlash: ['https://base.example/path1/'],
+        slash: ['https://base.example/']
       }
     );
   });
 
-  it('should ignore percent-encoded variants of ./, ../, or /', () => {
+  it('should treat percent-encoded variants of ./, ../, or / as non-URLs', () => {
```
domenic: So now we have bare specifiers in the output. This is not great for the ultimate, composed import map; ideally we'd give people a warning in such cases. I.e. it'd be ideal if bare specifiers in the first import map were discarded with a warning, bare specifiers in the second that did not map to anything from the first were discarded with a warning, etc.

I'm not sure whether that's feasible with the currently-envisioned architecture. (This is part of why I anticipated this being an even bigger change than it currently is.) I don't necessarily think we should demand it, since it's just about good warnings at parse time (instead of errors at use time), but we should weigh it.

Of course, if there's some nice way to make that work with minimal disruption, that'd be great as well.

reference-implementation/__tests__/parsing-addresses.js
```diff
@@ -152,25 +144,20 @@ describe('Absolute URL addresses', () => {
...
-        wss: []
+        about: ['about:good'],
+        blob: ['blob:good'],
+        data: ['data:good'],
+        file: ['file:///good'],
+        filesystem: ['filesystem:good'],
+        http: ['http://good/'],
+        https: ['https://good/'],
+        ftp: ['ftp://good/'],
+        import: ['import:bad'],
```
domenic: The loss of these warnings is also troubling. These aren't even bare specifiers... they're invalid absolute URL specifiers. Why did we lose the ability to reject them?

reference-implementation/__tests__/resolving-builtins.js
```diff
@@ -155,4 +168,16 @@ describe('Fallbacks with built-in module addresses', () => {
   it(`should fall back past "${NONE}"`, () => {
     expect(resolveUnderTest('none')).toMatchURL('https://example.com/app/none-fallback.mjs');
   });
+
+  it('should fall back for [built-in, built-in]', () => {
+    expect(resolveUnderTest('twoGoodBuiltins')).toMatchURL(BLANK);
+  });
+
+  it('should fail for [built-in, fetch scheme, fetch scheme]', () => {
```
domenic: These aren't quite correct. resolve() would return the first URL which was successfully fetched, most likely. The complexities there (e.g. making resolve() async or network dependent) are why we've kept this as unimplemented. It'd be better to stick to only supporting the [built-in, fetch scheme URL] pattern for now in the fallbacks.

reference-implementation/__tests__/resolving.js
```diff
@@ -63,6 +60,53 @@ describe('Unmapped', () => {
   });
 });
 
+describe('the empty string', () => {
+  it('should fail for an unmapped empty string', () => {
+    const resolveUnderTest = makeResolveUnderTest(`{}`);
+    console.warn = () => {};
```
domenic: Why is this necessary? Same below. It seems we should not get a warning if we're also getting a TypeError... But even if we do, we should test for the presence of the warning.

reference-implementation/lib/resolver.js
```diff
@@ -1,100 +1,60 @@
...
-      if (scopeImportsMatch !== null) {
-        return scopeImportsMatch;
-      }
-    }
-  }
-
-  const topLevelImportsMatch = resolveImportsMatch(normalizedSpecifier, parsedImportMap.imports);
-  if (topLevelImportsMatch !== null) {
-    return topLevelImportsMatch;
+exports.resolve = (specifier, parsedImportMap, scriptURLparameter) => {
```
domenic: If it's easy it would be nice to have this take a `URL` as its third argument, directly. (Bad JavaScript API design, but good for a reference implementation mirroring the spec.) I kind of thought it did previously...

reference-implementation/lib/utils.js
```diff
@@ -19,23 +19,56 @@ exports.tryURLParse = (string, baseURL) => {
 };
 
 exports.tryURLLikeSpecifierParse = (specifier, baseURL) => {
+  if (specifier === '') {
+    console.warn(`Invalid empty string specifier.`);
+    return { type: 'invalid' };
```
domenic: Maybe it'd be better to communicate the warning as a field in the return type, which can then be turned into a thrown TypeError, or console.warn()ed, by the caller? (Related to above comment about tests that set console.warn to a no-op.)

reference-implementation/__tests__/parsing-addresses.js
```diff
@@ -334,17 +316,17 @@ describe('Failing addresses: mismatched trailing slashes', () => {
 });
 
 describe('Other invalid addresses', () => {
-  it('should ignore unprefixed strings that are not absolute URLs', () => {
+  it('should treat unprefixed strings that are not absolute URLs as non-URLs', () => {
     for (const bad of ['bar', '\\bar', '~bar', '#bar', '?bar']) {
```
domenic: "bad" is probably no longer the right variable name here

reference-implementation/lib/resolver.js
```diff
@@ -1,100 +1,60 @@
...
-        // Enforced by parsing
-        assert(addresses[0].href.endsWith('/'));
-
-        // Cannot use URL resolution directly. E.g. "switch" relative to "std:elements/" is a
-        // parsing failure.
-        return new URL(addresses[0] + afterPrefix);
-      } else if (addresses.length === 2 &&
-                 addresses[0].protocol === BUILT_IN_MODULE_PROTOCOL &&
-                 addresses[1].protocol !== BUILT_IN_MODULE_PROTOCOL) {
+exports.getFallbacks = (normalizedSpecifier, parsedImportMap, resolutionContext) => {
```
domenic: resolutionContext -> scriptURL, maybe?

domenic (commented):
Thanks so much for the work on this, and sorry for the delay. This is a big change, but you've clearly gone out of your way to create a battery of tests, as well as do minor things like match the coding style. That makes it much easier.

I'm also heartened by the relative decrease in code size in resolver.js. Some of that may be incidental (e.g. facotring out some of the logic into tryURLLikeSpecifierParse, or removing our safeguards against some not-implemented cases, or using a functional style instead of a (more spec-translation-friendly) imperative. But the getFallbacks algorithm seems like an actual win.

The main changes I'd like to see is adding a few more of the tests in locations I mentioned inline. Also, I'd like @hiroshige-g to take a look at the new/modified tests, and see if he has any of his own suggestions for devious tests to add. For example, it seems possible to me that there aren't enough tests exploring the consequences of mapping bare specifiers twice, and how that might interact with scopes; I see good coverage for absolute URLs but not as much for bare specifiers. Also, I'd like to make sure we have coverage for nailing down an answer to #153.

Overall, it seems likely this can work. I continue to maintain that the result is a more complex mental model than the current specifiers -> URLs setup; I know you disagree. But I'm willing to accept that this increase in complexity in exchange for the compositional benefits you've worked to explain.

Finally, @hiroshige-g should also take a look at the new code logic, and judge if there are any problematic aspects. In particular I want to be sure that the new resolve() function is not too expensive in some way, since that is semi-hot. It seems to have the same algorithmic complexity over various inputs, but I haven't analyzed it rigorously, and it may benefit from some optimizations (e.g. there may be places you could reduce the number of O(n) iterations over a list).

guybedford (commented):
I haven't followed the exact details but the overall approach seems like a sensible model, and while adding complexity it doesn't seem to cross any lines to me.

I just thought I would mention that my main concern here would be that with this proposal, once a URL has been overridden, the original can never be referenced again.

So later import maps effectively have less control than earlier ones. Maybe that's a bug or a feature though! (interestingly it can actually be seen as a security feature to restrict builtins...).

reference-implementation/__tests__/parsing-addresses.js
bakkot: We should be able to give warnings for this and remove the specifiers from the output, but it would happen at composition time, not parse time. (You can conceptualize this as starting the document with an empty map and composing any that are present, in order.)

reference-implementation/__tests__/parsing-addresses.js
bakkot: Absolute URL specifiers with non-fetch schemes are treated as non-URLs, so in this case the specifier `import:bad` would be successfully *parsed* but not allowed to be the result of resolution. In [a separate test](https://github.com/WICG/import-maps/blob/ccaf0c062738a2c1b93470ffffa297ed3f251b33/reference-implementation/__tests__/resolving.js#L94-L108), we show how non-fetch-scheme-URL module specifiers like this one can be made to map to resolvable URLs without error.

After addressing [your earlier comment about stripping bare specifiers](https://github.com/WICG/import-maps/pull/167/files#r314491076), these would end up removed with a warning at composition time.

reference-implementation/__tests__/resolving-builtins.js
bakkot: For simplicity, this implementation assumes that all web resource fetches succeed (which seems to match the current tested behavior). We can artificially restrict the resolution algorithm to only accept fallback lists supported by the current implementation. Or we could use some different strategy to determine a success status.

reference-implementation/lib/resolver.js
bakkot: This is used in [the composer](https://github.com/WICG/import-maps/blob/ccaf0c062738a2c1b93470ffffa297ed3f251b33/reference-implementation/lib/composer.js#L23) to resolve fallbacks for a scope in the context of the existing map based on that scope (i.e., with the scope as the resolution context). In that case there's not yet a "script"; hence the name of this parameter.

bakkot:
@guybedford That's intentional. Otherwise the README's example of [denying access to a built-in module](https://github.com/WICG/import-maps/#denying-access-to-a-built-in-module) would be impossible; see #137 for the motivations behind this PR.

reference-implementation/__tests__/resolving.js
michaelficarra: The warning comes from parsing and the thrown `TypeError` comes from resolution. They're composed in `makeResolveUnderTest`. I'll assert that the warnings are present in the same way we do in `testWarningHandler`.

reference-implementation/__tests__/composition.js
bakkot: > should result in... just a bare specifier in the output, I think?

Yes, except that [this](https://github.com/WICG/import-maps/pull/167#discussion_r314064807) will change "the output is just a bare specifier" to the equivalent "the output is an empty array" (and a warning).

michaelficarra:
@domenic All comments addressed. I've left the commits separate to make it easier to review only the changes.

reference-implementation/__tests__/composition.js
michaelficarra: It was left over from when we had URL objects in the result, as a way to make it easier to write tests. Removed now.

reference-implementation/__tests__/parsing-addresses.js
bakkot: To spell this out a little more:

You can't give a warning for a bare specifier on the right-hand side _at parse time_, because that can be legal when adding an import map to a page which provides a mapping for that specifier. For example,

```html
<script type="importmap">
{
  "imports": {
    "moment": "/node_modules/moment/src/moment.js"
  }
}
</script>
<script type="importmap">
{
  "imports": {
    "m": "moment"
  }
}
</script>
```
is fine, despite the second map having the bare specifier `"moment"` on the RHS, because that `"moment"` is mapped to a real URL by the first map.

But you _can_ give a warning for a bare specifier on the RHS which is still present _after composition_, since maps only cascade to earlier maps, not later ones. And you can use this mechanism to give warnings for the first map on the page by acting as if every page starts out having the empty import map `{}` loaded already. For example,

```html
<script type="importmap">
{
  "imports": {
    "moment": "/node_modules/moment/src/moment.js",
    "lodash": "lodash/index.js"
  }
}
</script>
<script type="importmap">
{
  "imports": {
    "jquery": "javascript:bad",
    "l": "lodash"
  }
}
</script>
```

can give a warning for `"lodash"` when adding the first map and for both `"jquery"` and `"l"` when adding the second map, since you know when each map is added to the page that those point to bare specifiers. The PR has been updated with this behavior.

michaelficarra:
Ping @domenic. No rush, but I wanted to make sure you didn't miss that this is ready for review. Also, @hiroshige-g I believe we are waiting on a review from you as well.

domenic:
Thanks, this looks good. I have run out of time today, but I will work on porting the reference implementation updates to the spec text next week.

domenic:
I was trying to explain to @hiroshige-g the compositional mental model you explained to me offline. In particular I think your model was something like: authors expect the import maps to apply in reverse (?) order, eventually bottoming out at a "basic specifier", which gets an identity transform from specifier string -> URL space. But when trying to reproduce this on [the "virtualization patterns we expect to see in the wild"](https://github.com/WICG/import-maps/pull/167/files#diff-bc8e235e65b9e141f5db989de8f790d9R133) example, it didn't quite work that way.

In particular, consider a module that does `import "std:blank"` from inside `https://built-in-enhancement-2/`.

```js
const specifierFromSourceText = "std:blank";

const specifier = normalizeToURLString(specifierFromSourceText);
// -> "std:blank"

specifier = apply(map3, specifier);
// -> https://built-in-enhancement-3/

specifier = apply(map2, specifier);
// -> no change...

specifier = apply(map1, specifier);
// -> no change...

const url = parseToURL(specifier);
// -> https://built-in-enhancement-3/
```

whereas the test states the result should be `https://built-in-enhancement-1/`.

If the intuition is non-reversed, that fixes the result when the referring script's URL is inside `https://built-in-enhancement-2/`, but then we have a problem when working from inside `https://built-in-enhancement-1/`:


```js
const specifierFromSourceText = "std:blank";

const specifier = normalizeToURLString(specifierFromSourceText);
// -> "std:blank"

specifier = apply(map1, specifier);
// -> std:blank

specifier = apply(map2, specifier);
// -> https://built-in-enhancement-2/

specifier = apply(map3, specifier);
// -> no change...

const url = parseToURL(specifier);
// -> https://built-in-enhancement-2/
```

whereas the test says it should be `std:blank`.

I know we've implemented a different version of this (which I prefer), which involves creating a single composed import map. We should keep doing that. But I want to try to recapture the intuition you were explaining to me about how authors might think about it.

bakkot:
The simple mental model we explained makes sense only when all of the mappings are similarly specific. So, if all mappings are in the same scope or all at the top level, that *is* how composition works. It gets more complicated when adding an import map with less-specific scopes on top of an import map with more-specific scopes: in that case you would not want to apply a mapping from a less-specific scope when a more-specific scope gives a mapping for the module specifier of interest, since more specific scopes are considered to be higher precedence.

(At the time of our conversations with you, we hadn't yet considered the interactions of scope specificity with composition.)

domenic:
OK, the spec updates are almost done. (And the corresponding reference implementation tweaks.) What's remaining:

- [ ] Wait for https://github.com/whatwg/infra/pull/265 to get merged and reference that (currently there's a placeholder
- [ ] Add an example or two or three of concatenating import maps, and explain the mental model. This replaces the [previous, now-inaccurate example](https://wicg.github.io/import-maps/#example-merging). I would appreciate if @michaelficarra or @bakkot were willing to take that on (pushing to this branch).

hiroshige-g:
Thanks for clarification. So provided the scopes of different import maps don't interfere, then we can interpret the cascading as "apply import maps in-order, while scopes are exempted from further overriding by later import maps".

For example in https://github.com/WICG/import-maps/issues/137#issuecomment-512951464, the `"/kvs-v2-polyfill.mjs"` scope in the first import map is exempted from overriding by the second import map.

hiroshige-g:
I'm still highly concerned about the complexity (to implementation and to users) added by this.
Especially the change adds mixture of URL and non-URLs and parsing/serializing/reparsing URLs, which makes it harder to understand the import maps behavior.

I'd like to defer the final decision at least after we pass Chromium-internal reviews (preparing...) on the basic parts of import maps that are already reflected in the current draft spec.

bakkot:
@domenic 

> Add an example or two or three of concatenating import maps, and explain the mental model. This replaces the previous, now-inaccurate example. I would appreciate if @michaelficarra or @bakkot were willing to take that on (pushing to this branch).

Will do. I'll try to get to this early next week.

@hiroshige-g

> So provided the scopes of different import maps don't interfere, then we can interpret the cascading as "apply import maps in-order, while scopes are exempted from further overriding by later import maps".

In reverse order, but yes.

An (equivalent) way of describing it is "find the most specific applicable scope which provides a mapping for the module specifier being evaluated, across all maps, and apply maps in reverse order starting from the map with that scope".

Or, if you are thinking about about composition directly, you can (again equivalently) say "when merging in a new map, the keys on its RHS are evaluated in the context of the existing map, just as any other module specifier would be. Scoped mappings are evaluated as if in that scope."

hiroshige-g:
Another example of breaking the "apply import maps in-order" semantics:
In the current PR, IIUC

```
<script type="importmap">
{
    "imports": {
      "bar/": "https://bar.example.com/",
      "bar/baz/": "https://baz.example.com/"
     }
}
</script>
<script type="importmap">
{
    "imports": { "foo/": "bar/" }
}
</script>
```

is combined into

```
<script type="importmap">
{
    "imports": { "foo/": "https://bar.example.com/" }
}
</script>
```

and thus `foo/baz/a.js` is mapped into `https://bar.example.com/baz.a.js`,
not `foo/baz/a.js` -> `bar/baz/a.js` -> `https://baz.example.com/a.js`.

This is inevitable though...

hiroshige-g:
Er, can this PR already result in exponential-sized output?
For example:

```
<script type="importmap">
{
    "imports": {
      "a1": "https://example.com/a1",
      "a2": "https://example.com/a2"
     }
}
</script>
<script type="importmap">
{
    "imports": {
      "b1": ["a1", "a2"],
      "b2": ["a1", "a2"]
     }
}
</script>
<script type="importmap">
{
    "imports": {
      "c1": ["b1", "b2"],
      "c2": ["b1", "b2"]
     }
}
...
</script>
```

result in `O(2^n)` addresses for `n` import maps?

bakkot:
> Er, can this PR already result in exponential-sized output?

Hm. As implemented, yes, because there's nothing forbidding having the same specifier appear multiple times in a fallback list (even within a single map). It doesn't really make sense to allow that, though. If fallback lists are deduplicated during composition I believe that problem goes away.

hiroshige-g:
I agree that there are no valid use cases for exponentially-exploding cases. But even with deduplication, trailing-slash matching can result in exponential-sized output.

```
<script type="importmap">
{
    "imports": {
      "a1/": "https://example.com/1/",
      "a2/": "https://example.com/2/"
     }
}
</script>
<script type="importmap">
{
    "imports": {
      "b1/": ["a1/1/", "a2/1/"],
      "b2/": ["a1/2/", "a2/2/"]
     }
}
</script>
<script type="importmap">
{
    "imports": {
      "c1": ["b1", "b2"],
      "c2": ["b1", "b2"]
     }
}
...
</script>
```

bakkot:
Ah, yes, I keep forgetting about package-prefix mappings.

In this case that seems to be what the user is asking for, though, so I'm not sure to what extent it should be regarded as a problem. In an actual implementation it could be represented in memory without the exponential overhead, if that's the worry.

On the other hand, would browsers in practice cap how many times they attempted to fetch a given module? If so, it seems reasonable to cap the number of fallbacks to that number.

guybedford:
Might this be an argument for treating cascade resolution as actually performing the fallback selection?

lemanschik:
just my 5 cent doesen't that overall add a big extra problem for that we need to create additional tooling i see we shifting the problems around we change the place where they happen this should when get done be done only inside a serviceWorker that that it gives a JS static analyzeable way to audit track what does resolve to what.

domenic:
We went a different direction for multiple import maps support.

PR #168, opened 2019-07-25
Include Appendix A media type registration

guybedford:
This includes the media type registration as an Appendix A in the specification text, according to the process for W3C registrations outlined at https://www.w3.org/2002/06/registering-mediatype2014.html.

I have included myself as the contact for further information for now, but would be happy to change this if another contact would like to be listed - I just didn't want to incorrectly assume this responsibility for anyone else.


spec.bs
```diff
@@ -569,3 +569,50 @@ Alternatively, we can add the snippet at the beginning of the following HTML spe
...
+:: N/A
+: Magic number(s)
+:: N/A
+: File extension(s):
+:: .importmap
+: Macintosh file type code:
+:: N/A
+: Object Identifiers:
+:: N/A
+: Person & email address to contact for further information:
```
yoavweiss: @wseltzer / @plehegar - Thoughts on the right contact person here?

spec.bs
```diff
@@ -569,3 +569,50 @@ Alternatively, we can add the snippet at the beginning of the following HTML spe
...
+: Type name:
+:: application
+: Subtype name:
+:: importmap+json
+: Required parameters:
+:: N/A
+: Optional parameters:
+:: N/A
+: Encoding considerations:
+:: 8bit
```
domenic:
```suggestion
:: 8bit (always UTF-8)
```

following the example of https://html.spec.whatwg.org/#text/cache-manifest

spec.bs
```diff
@@ -569,3 +569,50 @@ Alternatively, we can add the snippet at the beginning of the following HTML spe
...
+: Security considerations:
+:: Since Import Maps have the ability to direct which module resolutions are to be provided to all ECMAScript module imports of a given ECMAScript environment, control of the Import Map should be considered to be execution-level application access. Integration with existing policy systems, including for example CORS and CSP, can be used to mitigate and restrict unwanted target URLs from being executed. In addition, the specification states that only those Import Maps served to browsers with the `application/importmap+json` MIME type will be executed.
```
domenic:
```suggestion
:: Since import maps have the ability to direct which module resolutions are to be provided to all JavaScript module imports of a given JavaScript environment, control of the import map should be considered to be execution-level application access. Integration with existing policy systems, including for example CORS and CSP, can be used to mitigate and restrict unwanted target URLs from being executed. In addition, the specification states that only those import maps served to browsers with the `application/importmap+json` MIME type will be executed.
```

spec.bs
```diff
@@ -569,3 +569,50 @@ Alternatively, we can add the snippet at the beginning of the following HTML spe
...
+: Interoperability considerations:
+:: Backwards compatibility will be a necessity for any new features added to the format, and some handling for this has been incorporated into the design of the specification.
```
domenic:
```suggestion
:: Backwards compatibility will be a necessity for any new features added to the format, and handling for this has been incorporated into the design of the specification.
```

spec.bs
```diff
@@ -569,3 +569,50 @@ Alternatively, we can add the snippet at the beginning of the following HTML spe
...
+: Published specification:
+:: https://wicg.github.io/import-maps/
+: Applications that use this media type:
+:: This is a browser-specific MIME type but may also apply to other ECMAScript environments.
```
domenic:
```suggestion
:: This is a browser-specific MIME type but may also apply to other JavaScript environments.
```

spec.bs
```diff
@@ -569,3 +569,50 @@ Alternatively, we can add the snippet at the beginning of the following HTML spe
...
+: Fragment identifier considerations:
+:: N/A
+: Restrictions on usage:
+:: N/A
```
domenic:
```suggestion
:: No restrictions apply.
```

spec.bs
```diff
@@ -569,3 +569,50 @@ Alternatively, we can add the snippet at the beginning of the following HTML spe
...
+: Additional information:
+: Deprecated alias names for this type:
```
domenic: These should be indented (cf. https://html.spec.whatwg.org/#text/cache-manifest)

spec.bs
```diff
@@ -569,3 +569,50 @@ Alternatively, we can add the snippet at the beginning of the following HTML spe
...
+:: N/A
+: Magic number(s)
+:: N/A
+: File extension(s):
+:: .importmap
```
domenic:
```suggestion
:: "`importmap`"
```

spec.bs
```diff
@@ -569,3 +569,50 @@ Alternatively, we can add the snippet at the beginning of the following HTML spe
...
+: Security considerations:
+:: Since Import Maps have the ability to direct which module resolutions are to be provided to all ECMAScript module imports of a given ECMAScript environment, control of the Import Map should be considered to be execution-level application access. Integration with existing policy systems, including for example CORS and CSP, can be used to mitigate and restrict unwanted target URLs from being executed. In addition, the specification states that only those Import Maps served to browsers with the `application/importmap+json` MIME type will be executed.
+: Interoperability considerations:
+:: Backwards compatibility will be a necessity for any new features added to the format, and some handling for this has been incorporated into the design of the specification.
+: Published specification:
+:: https://wicg.github.io/import-maps/
+: Applications that use this media type:
+:: This is a browser-specific MIME type but may also apply to other ECMAScript environments.
+: Fragment identifier considerations:
+:: N/A
```
domenic: Remove N/A sections?

spec.bs
```diff
@@ -569,3 +569,50 @@ Alternatively, we can add the snippet at the beginning of the following HTML spe
...
+: File extension(s):
+:: .importmap
+: Macintosh file type code:
+:: N/A
+: Object Identifiers:
+:: N/A
+: Person & email address to contact for further information:
+:: Guy Bedford (guybedford@gmail.com).
+: Intended usage:
+:: COMMON
```
domenic:
```suggestion
:: Common
```

spec.bs
```diff
@@ -569,3 +569,50 @@ Alternatively, we can add the snippet at the beginning of the following HTML spe
...
+: Restrictions on usage:
```
domenic: Appears twice

spec.bs
```diff
@@ -569,3 +569,50 @@ Alternatively, we can add the snippet at the beginning of the following HTML spe
...
+:: N/A
+: Change controller:
+:: W3C
```
domenic: I'm not sure how much sense this makes given that import maps will become part of the WHATWG HTML specification.

RFC 6838 doesn't say much about this field. Perhaps just your name would work?

spec.bs
```diff
@@ -569,3 +569,50 @@ Alternatively, we can add the snippet at the beginning of the following HTML spe
...
+: Magic number(s)
+:: N/A
+: File extension(s):
+:: .importmap
+: Macintosh file type code:
+:: N/A
+: Object Identifiers:
+:: N/A
+: Person & email address to contact for further information:
+:: Guy Bedford (guybedford@gmail.com).
```
domenic:
```suggestion
:: Guy Bedford &lt;guybedford@gmail.com>.
```

spec.bs
```diff
@@ -569,3 +569,50 @@ Alternatively, we can add the snippet at the beginning of the following HTML spe
 </div>
 
 <p class="note">[=Validate a module script URL=] is applied to all module URLs before they start loading, even in paths where [=resolve a module specifier=] and import maps are not applied (e.g. `<script src="..." type="module">`).</p>
+
+<hr>
+<h2 id="internet-media-type-registration">Appendix A: Internet Media Type Registration</h2>
```
domenic:
```suggestion
<h2 id="internet-media-type-registration">Appendix A: MIME type registration</h2>
```

spec.bs
```diff
@@ -569,3 +569,50 @@ Alternatively, we can add the snippet at the beginning of the following HTML spe
...
+
+This appendix provides the provisional registration of the Internet Media Type `application/importmap+json` in accordance with [[RFC6838]].
```
domenic:
```suggestion
This appendix provides the provisional registration of the [=MIME type=] <dfn export>`application/importmap+json`</dfn> in accordance with <cite>Media Type Specifications and Registration Procedures</cite>. [[RFC6838]]
```

spec.bs
guybedford: Any ideas how I can do this with this syntax?

domenic: Not sure; I'm more of a HTML person myself.

spec.bs
guybedford: Since these are part of the template, we need to leave them in explicitly.

guybedford: Since these are part of the registration template, we should strictly include them.

guybedford:
I've included all the feedback here now. Note that for both the required parameters and optional parameters, referring to the JSON specification is effectively referring to the `N/A` values. Using an `N/A` field is very specifically requested to be provided for non-applicable fields in the template, so directly providing this value seems better. I did refer to the JSON spec for the Mac filetype code though, being the only one of these references that is not N/A.

Let me know if there is any further feedback to merge further.

guybedford:
The CI error here seems to be an unrelated issue with the lockfile in the reference implementation.

spec.bs
plehegar: I don't recommend putting individual email address there since it needs to survive individuals moving on but I have no objection if you folks feels otherwise. At the end of the day, the best way to reach out is through GitHub as far as I know.
Basically, we'd need an email address that someone can see and redirect to the right person if needed. Some ideas, with various pros/cons:
* Domenic Denicola public-wicg@w3.org
* Domenic Denicola www-archive@w3.org

PR #180, opened 2019-10-02
Remove fetch scheme restriction in `parse a URL-like import specifier`

### LINKED ISSUE #141: Behavior changes in specifier resolution without import maps
#resolve-a-module-specifier fails for non-fetch scheme URLs because of the check at Step 4 of #parse-a-url-like-import-specifier, while previously #resolve-a-module-specifier itself succeeded (subsequent fetch would still fail though).
This slightly affects the behavior of pages even without import maps.

For example,

```
<script type="module">
import './fetch-failure.mjs'; // The fetch will fail.
import 'nonfetchscheme:foobar';
</script>
```

The current HTML spec, all the specifier resolution are successful, and two fetches to `fetch-failure.mjs` and `nonfetchscheme:foobar` are made and failed.
Fetch failure of `fetch-failure.mjs` is reported to error console.

In the current import map spec, the specifier resolution of `nonfetchscheme:foobar` fails, and thus creating the top-level inline module script also fails. In this case, no network request to `fetch-failure.mjs` is sent to the server, and creation error of the inline script is reported to error console.

I don't expect this change is important, as I don't expect stable prodution systems contain such non-fetch scheme URLs and errors.
Probably it's better to first land a HTML spec change to make this behavior change (that can be done independently from import maps), and then land import map changes (without behavior changes to non-import-map cases).

hiroshige-g:
This allows non-fetch-scheme URLs as

- import map's keys (as URLs rather than as bare specifiers),
- import map's values, and
- the result of `resolve a module specifier` algorithm.

Fixes #141.

***
<a href="https://pr-preview.s3.amazonaws.com/hiroshige-g/import-maps/pull/180.html" title="Last updated on Oct 6, 2019, 8:28 AM UTC (53c0842)">Preview</a> | <a href="https://pr-preview.s3.amazonaws.com/WICG/import-maps/180/eacf4ea...hiroshige-g:53c0842.html" title="Last updated on Oct 6, 2019, 8:28 AM UTC (53c0842)">Diff</a>

### DIFF
--- a/reference-implementation/__tests__/parsing-addresses.js
+++ b/reference-implementation/__tests__/parsing-addresses.js
@@ -105,14 +105,13 @@ describe('Absolute URL addresses', () => {
         filesystem: expect.toMatchURL('filesystem:good'),
         http: expect.toMatchURL('http://good/'),
         https: expect.toMatchURL('https://good/'),
-        ftp: expect.toMatchURL('ftp://good/')
+        ftp: expect.toMatchURL('ftp://good/'),
+        import: expect.toMatchURL('import:bad'),
+        javascript: expect.toMatchURL('javascript:bad'),
+        mailto: expect.toMatchURL('mailto:bad'),
+        wss: expect.toMatchURL('wss://bad/')
       },
-      [
-        `Invalid address "import:bad" for the specifier key "import".`,
-        `Invalid address "mailto:bad" for the specifier key "mailto".`,
-        `Invalid address "javascript:bad" for the specifier key "javascript".`,
-        `Invalid address "wss:bad" for the specifier key "wss".`
-      ]
+      []
     );
   });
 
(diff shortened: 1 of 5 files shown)


marcoscaceres:

marcoscaceres:
Marking as non substantive as it's a tiny change... 

@hiroshige-g, if you'd like to continue continue contributing we'd encourage you to join the WICG! Simply go to https://wicg.io -> Click Join -> and then click the join on the page you get redirected to.  If you've already joined, you need to link your GitHub account to your W3C account on w3.org.

hiroshige-g:
Thanks for instruction! I've submitted a join request.

guybedford:
If the fetch scheme filter doesn't add anything, this seems sensible to me.

PR #182, opened 2019-10-03
Sort keys in the reverse order of `code unit less than`

### LINKED ISSUE #181: longer-or-code-unit-less-than is not a total order
The order used to sort scope keys etc., i.e.
https://wicg.github.io/import-maps/#longer-or-code-unit-less-than
> if a’s length is greater than b’s length, or if a is code unit less than b.

is not a total order. For example,

- `"A"` is longer-or-code-unit-less-than `"AAA"`
- `"AAA"` is longer-or-code-unit-less-than `"A"`

Probably this should be something like

> if a’s length is greater than b’s length, or if a’s length is equal to b’s length and a is code unit less than b.

or simply

> if b is code unit less than a.

Probably the latter one is better, because it's simple to understand and implement.
I'll create a PR shortly.

(Memo: `code unit less than` also seems to have a spec bug)

hiroshige-g:
Fixes #181.

***
<a href="https://pr-preview.s3.amazonaws.com/hiroshige-g/import-maps/pull/182.html" title="Last updated on Oct 15, 2019, 6:47 AM UTC (cb51eb7)">Preview</a> | <a href="https://pr-preview.s3.amazonaws.com/WICG/import-maps/182/eacf4ea...hiroshige-g:cb51eb7.html" title="Last updated on Oct 15, 2019, 6:47 AM UTC (cb51eb7)">Diff</a>

### DIFF
--- a/spec.bs
+++ b/spec.bs
@@ -336,7 +336,7 @@ To <dfn>register an import map</dfn> given an {{HTMLScriptElement}} |element|:
   1. Return the result of [=map/sorting=] |normalized|, with an entry |a| being less than an entry |b| if |b|'s [=map/key=] is [=code unit less than=] |a|'s [=map/key=].
 </div>
 
-<div class="note">We sort keys/scopes in reverse order, to put `"foo/bar/"` before `"foo/"` so that `"foo/bar/"` has a higher priority than `"foo/"`.</div>
+<p class="note">We sort keys/scopes in reverse order, to put `"foo/bar/"` before `"foo/"` so that `"foo/bar/"` has a higher priority than `"foo/"`.</p>
 
 <div algorithm>
   To <dfn lt="normalize a specifier key|normalizing a specifier key">normalize a specifier key</dfn>, given a [=string=] |specifierKey| and a [=URL=] |baseURL|:
--- a/spec.bs
+++ b/spec.bs
@@ -315,7 +315,7 @@ To <dfn>register an import map</dfn> given an {{HTMLScriptElement}} |element|:
       1. [=Report a warning to the console=] that an invalid address was given for the specifier key |specifierKey|; since |specifierKey| ended in a slash, so must the address.
       1. [=Continue=].
     1. Set |normalized|[|specifierKey|] to |addressURL|.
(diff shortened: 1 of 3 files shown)


spec.bs
```diff
@@ -315,7 +315,7 @@ To <dfn>register an import map</dfn> given an {{HTMLScriptElement}} |element|:
       1. [=Report a warning to the console=] that an invalid address was given for the specifier key |specifierKey|; since |specifierKey| ended in a slash, so must the address.
       1. [=Continue=].
     1. Set |normalized|[|specifierKey|] to |addressURL|.
-  1. Return the result of [=map/sorting=] |normalized|, with an entry |a| being less than an entry |b| if |a|'s [=map/key=] is [=longer or code unit less than=] |b|'s [=map/key=].
+  1. Return the result of [=map/sorting=] |normalized|, with an entry |a| being less than an entry |b| if |b|'s [=map/key=] is [=code unit less than=] |a|'s [=map/key=].
```
guybedford: Are you making the assumption that a code unit outside the end of the string is taken to be zero and thus always less than any other code unit? That seems like something that shouldn't necessarily be an implicit assumption to me.

hiroshige-g: https://infra.spec.whatwg.org/#code-unit-less-than explicitly handles prefixes (Steps 1 and 2) and we don't have to care about code unit outside the end of strings.

guybedford: Ah perfect, glad to see this is well-defined.

marcoscaceres:
@hiroshige-g please be sure to ping me once you've joined the group and I can click the magic buttons to unblock this. Thanks again!

hiroshige-g:
@marcoscaceres Thanks for handling! It seems I'm now a WICG participant.

marcoscaceres:
Ok, all green now. Good to merge!

hiroshige-g:
> unable to create a test that failed before this change and passed after it.

I think this is fine, because:
In the reference implementation before this PR the sort order was `if the length is greater, or equal and a greater code unit`, not `if the length is greater, or a greater code unit` as specified, and thus the resolution results by the ref impl were already correct.

PR #183, opened 2019-10-03
Add test case for overriding across scopes

hiroshige-g:


marcoscaceres:

PR #196, opened 2020-01-02
Add fairly important "a"

jkrems:
I assume this is a typo fix and this is what was meant - fallbacks may make the resolution process *a*synchronous.


domenic:
Looks good, although I guess we should wait for one of the WICG admins to mark your contribution as "non-substantive"

dontcallmedom:

PR #204, opened 2020-01-29
Migrate to JSON-based Jest tests

hiroshige-g:
In order to share common structure with Web Platform Tests #170,
this PR migrates hard-coded Jest tests into
JSON test metadata (`__tests__/json/*.json`, which are shared with
https://github.com/web-platform-tests/wpt/tree/master/import-maps/common/resources)
and test runners (`__tests__/helpers/`, which are Jest-specific).

PR #205, opened 2020-01-29
Make any errors (invalid URLs etc) throw errors in resolution

### LINKED ISSUE #184: Clarification: #176 and invalid URLs
#176 changed the behavior around invalid URLs. i.e.

`{"imports": { "https://example.com/foo": "https://:invalid:url" }}`

is equal to

- `{"imports": {}}` after https://github.com/WICG/import-maps/pull/176, i.e. `"https://example.com/foo"` remains unmapped (i.e. import maps can't be used to force-reject specifier resolution), while
- `{"imports": { "https://example.com/foo": null }}` before https://github.com/WICG/import-maps/pull/176, i.e. resolving `"https://example.com/foo"` fails.

What was the reason behind this change?

Does this mean that, after #176, when we re-enable fallback/built-in modules support again:

1. `{"imports": { "https://example.com/foo": "https://:invalid:url" }}` is equal to `{"imports": {}}` as a direct result of the change #176,
2. `{"imports": { "https://example.com/foo": ["https://:invalid:url"] }}` is equal to `{"imports": {}}` for keeping equivalence between `"https://:invalid:url"` and `["https://:invalid:url"]` that held before #176,
3. `{"imports": { "https://example.com/foo": [] }}` is equal to `{"imports": {}}`,
4.  `{"imports": { "https://example.com/foo": null }}` is different from `{"imports": { "https://example.com/foo": [] }}`

?

Context: Chromium will keep fallback implementation behind built-in modules flag, even after #176, and thus I'd like to clarify the background/implication of #176 to the fallback mechanism.
I'm neutral about the new/old behaviors.

hiroshige-g:
This PR all errors (found either at parsing or resolution time) throw errors at resolution.

Before this PR, errors found at parsing time, namely:

- Non-string values, including `null`, `[]`, etc.
- Invalid URLs
- Entries with keys with trailing `/` and values without trailing `/`

were ignored, allowing fallback to next candidates at resolution time, and thus can't be used as a definitive way to block particular resolution.
After this PR, such entries are kept in specifier maps as entries with `null` value, and cause resolutions fail (without further fallback) by throwing errors.
For thus purpose, this PR changes the type of values of specifier maps to `URL` or `null`.

Also, before this PR, relative URL resolution failures at prefix matching in `#resolve-a-module-specifier` didn't allow fallback to less specific prefixes while allowing fallback to less specific scopes.
This PR makes such errors throw errors, disallowing fallback to less specific prefixes and scopes.

This PR unifies resolution failures into one: throwing an error.
After this PR, all kind of errors throw at resolution time and definitively blocking resolution.
Fallback to less specific scopes occur only when there are no matching entries in a scope.

Fixes #184.

***
<a href="https://pr-preview.s3.amazonaws.com/hiroshige-g/import-maps/pull/205.html" title="Last updated on Feb 6, 2020, 9:54 PM UTC (630066c)">Preview</a> | <a href="https://pr-preview.s3.amazonaws.com/WICG/import-maps/205/6cb173d...hiroshige-g:630066c.html" title="Last updated on Feb 6, 2020, 9:54 PM UTC (630066c)">Diff</a>

### DIFF
--- a/reference-implementation/lib/resolver.js
+++ b/reference-implementation/lib/resolver.js
@@ -34,11 +34,19 @@ function resolveImportsMatch(normalizedSpecifier, specifierMap) {
   for (const [specifierKey, address] of Object.entries(specifierMap)) {
     // Exact-match case
     if (specifierKey === normalizedSpecifier) {
-      return address;
+      if (address === null) {
+        throw new TypeError(`Blocked by a null entry for "${specifierKey}"`);
+      } else {
+        return address;
+      }
     }
 
     // Package prefix-match case
-    if (specifierKey.endsWith('/') && normalizedSpecifier.startsWith(specifierKey)) {
+    else if (specifierKey.endsWith('/') && normalizedSpecifier.startsWith(specifierKey)) {
+      if (address === null) {
+        throw new TypeError(`Blocked by a null entry for "${specifierKey}"`);
+      }
+
       const afterPrefix = normalizedSpecifier.substring(specifierKey.length);
 
       // Enforced by parsing
@@ -48,7 +56,7 @@ function resolveImportsMatch(normalizedSpecifier, specifierMap) {
 
       // This code looks stupid but it follows the spec more exactly and also gives code coverage a chance to shine.
       if (url === null) {
-        return null;
+        throw new TypeError(`Failed to resolve prefix-match relative URL for "${specifierKey}"`);
       }
       return url;
     }
--- a/reference-implementation/lib/resolver.js
+++ b/reference-implementation/lib/resolver.js
@@ -32,17 +32,24 @@ exports.resolve = (specifier, parsedImportMap, scriptURL) => {
 
 function resolveImportsMatch(normalizedSpecifier, specifierMap) {
   for (const [specifierKey, resolutionResult] of Object.entries(specifierMap)) {
+    // Exact-match case
     if (specifierKey === normalizedSpecifier) {
-      // Exact-match case
       if (resolutionResult === null) {
         throw new TypeError(`Blocked by a null entry for "${specifierKey}"`);
       }
+
+      assert(resolutionResult instanceof URL);
+
       return resolutionResult;
-    } else if (specifierKey.endsWith('/') && normalizedSpecifier.startsWith(specifierKey)) {
-      // Package prefix-match case
+    }
+
+    // Package prefix-match case
+    if (specifierKey.endsWith('/') && normalizedSpecifier.startsWith(specifierKey)) {
       if (resolutionResult === null) {
         throw new TypeError(`Blocked by a null entry for "${specifierKey}"`);
       }
+
+      assert(resolutionResult instanceof URL);
 
       const afterPrefix = normalizedSpecifier.substring(specifierKey.length);
 
@@ -54,6 +61,9 @@ function resolveImportsMatch(normalizedSpecifier, specifierMap) {
       if (url === null) {
         throw new TypeError(`Failed to resolve prefix-match relative URL for "${specifierKey}"`);
       }
+
+      assert(url instanceof URL);
+
       return url;
     }
   }
--- a/reference-implementation/lib/resolver.js
+++ b/reference-implementation/lib/resolver.js
@@ -31,29 +31,28 @@ exports.resolve = (specifier, parsedImportMap, scriptURL) => {
 };
 
 function resolveImportsMatch(normalizedSpecifier, specifierMap) {
-  for (const [specifierKey, address] of Object.entries(specifierMap)) {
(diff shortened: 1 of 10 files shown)


hiroshige-g:
I'll create an associated test update PR later (due to dependency to #204; drafted locally).

reference-implementation/lib/resolver.js
```diff
@@ -48,7 +56,7 @@ function resolveImportsMatch(normalizedSpecifier, specifierMap) {
 
       // This code looks stupid but it follows the spec more exactly and also gives code coverage a chance to shine.
```
domenic: You can delete this comment now :)

hiroshige-g:
Added/updated ref impl and tests. Ready for review.

spec.bs
```diff
@@ -371,13 +371,11 @@ To <dfn>register an import map</dfn> given an {{HTMLScriptElement}} |element|:
...
-For each entry, the result is one of the following:
+  During [=resolve a module specifier|resolving a module specifier=], the following algorithms check candidate entries of [=specifier maps=], from most-specific to least-specific scopes (falling back to top-level "`imports`"), and from most-specific to least-specific prefixes. For each candidate, the result is one of the following:
 
-  - Successfully resolves a specifier to a [=URL=].
-    This makes the [=resolve a module specifier=] algorithm return immediately successfully with the [=URL=].
-  - Throws an error.
-    This makes the [=resolve a module specifier=] algorithm fail immediately, i.e. throw the error, without any further fallbacks.
+  - Successfully resolves a specifier to a [=URL=]. This makes the [=resolve a module specifier=] algorithm immediately return that [=URL=].
+  - Throws an error. This makes the [=resolve a module specifier=] algorithm rethrow the error, without any further fallbacks.
+  - Fail to resolve, without an error. In this case the algorithm moves on to the next candidate.
```
hiroshige-g: nit: I'm wondering better labels...
IIUC the third case corresponds to Step 1.3 of #resolve-an-imports-match, right?
"Throws an error" and "Fail to resolve" sound similarly and thus the difference between them might be unclear.
Alternatives I came up with are:

- "Throws an error" and "Continue"
- "Fail to resolve, throwing an error" and "Fail to resolve, without throwing an error"
  (Step 1.3 is just handling non-matching specifier keys, which doesn't look like failing something though)

Any idea?

hiroshige-g:
> The reference implementation

> should we align the spec with the reference implementation or the reference implementation with the spec.

One point I'm aware (and I was not sure) is Step 1.4-1.5 of #resolve-an-imports-match.
In the spec, all control paths sets `resolutionResult` and then Step 1.4-1.5 handle `resolutionResult` in a centralized fashion,
while in the ref impl each cases immediately throws etc.

Pros of ref impl style:

- Each case can provide more detailed error messages attached with errors, depending on contexts.

Cons of ref impl style:

- It's not clear that there are only three kind of resolution results for each entry, because there are no centralized result handling (like Step 1.4-1.5 in the spec).

Probably I should further refactor both the spec and implementation. I will try tomorow.

Or are there other diffs?

spec.bs
domenic: I was thinking it it corresponds to both step 1.3 and step 2.

Hmm maybe

- Resolves the specifier to null. This terminates the resolve a module specifier algorithm with a thrown exception.
- Does not resolve the specifier. In this case the algorithm moves on to the next candidate.

?

domenic:
> It's not clear that there are only three kind of resolution results for each entry

I don't think this is super clear even from the spec style. (The note attempts to clarify it, but as seen from our above discussion, the exact correspondence between those three cases and different steps in the algorithms is not 1:1.) So, I think I lean toward the reference implementation style.

spec.bs
hiroshige-g: "Resolves the specifier to URL", "Resolves the specifier to null" and "Does not resolve the specifier" looked great to me in terms of symmetry, but on second thought "Resolves to null" might be confusing with "Return null" in `resolve an imports match` (which is actually means "Does not resolve"), and might be a little inconsistent with `resolve a module specifier` interface (which throws exceptions rather than returning null).

"Does not resolve the specifier" still looks good to me, so how about something like:

- "Resolves the specifier to URL", "Throws an error" and "Does not resolve the specifier"
- "Resolves the specifier to URL", "Resolves the specifier to an error" and "Does not resolve the specifier"

?

hiroshige-g:
+1 to the reference implementation style.
I tried to refactor the spec to clarify the invariant, but it didn't turned successful (needed more plumbings, subalgorithms etc.).

Preparing a commit...

hiroshige-g:
Pushed. WDYT?

hiroshige-g:
Looks good!

Issue #209, opened 2020-02-22
Proposal: Solving the waterfall problem with depcache


guybedford:
I'd like to propose a new `"depcache"` field in the import map, as an optimal solution to the waterfall problem for production optimization.

The core idea is to provide the ability for tools that generate import maps to also generate the metadata of the module graph as a sort of graph cache (dependency cache) at the same time - a map from URLs to their list of dependency specifiers.

With a populated depcache, as soon as a module is fetched, the depcache can be consulted, and the corresponding cached list of module dependencies preloaded immediately.

For example:

```json
{
  "imports": {
    "main": "/main.js",
    "lazy-feature": "/lazy-feature.js,
    "dep1": "/lib/dep1.js",
    "dep2": "/lib/dep2.js"
  },
  "depcache": {
    "/main.js": ["dep"],
    "/lazy-feature.js": ["dep1", "dep2"]
  }
}
```

Say the main app loads `"main"` on startup. The depcache allows us to know that we need to also load `/lib/dep1.js` immediately. Both requests are made in parallel, possibly from the cache, the app thus working with zero latency waterfalls applying.

Later on a dynamic `import('lazy-feature')` is executed. At this point, the depcache can again be consulted, to see that `/lazy-feature.js` will import both `dep1` and `dep2`. Since `dep1` is already in the module map we do not need to send a new request, but then immediately send out just two requests - one for `/lazy-feature.js` and one for `/lib/dep2.js`, again getting lazy loading with full immediate parallel requests, without any duplication and supporting far-future caching.

Note that the unresolved dependency is included in the depcache. This allows the import map to remain the full source of truth for dependency resolution, and the depcache doesn't go stale despite resolutions changing.

The alternative as mentioned in the readme of this spec is a more general preloading spec. I have discussed with various spec authors and implementors the more general preload spec over the past year or two, and have found there to be a lack of interest - mostly I think because for most web resources, 2-3 round trips is the maximum due to the nature of HTML / CSS development. I would argue that modules are actually unique in having the problem of potentially N levels deep dependency trees, where N can be over 10, thus the latency problem between these depths does seem to me to be truly unique to modules.

This depcache proposal seems to me the simplest path forward right now to solve this optimization problem in a fully optimal way, given that a more general preload manifest is not getting traction, but I'm hoping by posting both this proposal and https://github.com/WICG/import-maps/issues/208, we can drive these conversations forward to ensure that this production optimization solution is tackled, as it really needs to be now for modules.

[...]

Issue #208, opened 2020-02-15
Driving work on a preload spec


guybedford:
I would like to reopen the discussion about preload metadata as mentioned in https://github.com/WICG/import-maps#supplying-out-of-band-metadata-for-each-module and previously discussed in https://github.com/WICG/import-maps/issues/21.

From the comment in the README, the allusion is towards a general manifest here, as purely an illustration, something like:

```html
<script type="preload-manifest">
{
  "https://site.com/path/to/module.js": {
    "integrity": "...",
    "preloads": [{
      "href": "./x.js",
      "rel": "modulepreload"
    }]
  },
  "https://site.com/path/to/x.js": {
    "integrity": "...",
    "credentials": "include",
    // eg dynamic import, alternatively some kind of priority system
    "lazyPreloads": [{
      "href": "import:dep",
      "rel": "modulepreload"
    }]
  }
}
</script>
```

Note that the "integrity" and other preload information does not associate with the dependency entry itself, but rather associates with the top-level entry. This is to avoid repeated information (and possibly conflicting information), in line with such a manifest being able to be a source of truth.

In trying to specify something along these lines I'm hitting some conceptual issues, and would value some feedback on how such a thing might work:

1. Strictly speaking, having purely a URL as the key in the manifest is not enough information for a general manifest. For example, `fetch('https://site.com/path/to/module.js')` should probably not trigger the preloading. That is, the preloading information needs to apply alongside the type of interpretation, currently captured by the `"rel"` or `"as"` attributes. Since the same resource may have multiple representations / loading profiles depending on _how_ it is loaded. It should probably be something more like:

```html
<script type="preload-manifest">
[
  {
    "url": "https://site.com/path/to/module.js",
    "rel": "modulepreload",
    "integrity": "...",
    "preloads": [...]
  },
]
</script>
```
This way each entry can participate in the preload system corresponding to its correct interpretation directly. At this point the manifest starts to look pretty convoluted though. Is there a solution to this problem?

2. Dependencies like `import 'dep'` can only use an `href` attribute to reference their dependency that is resolved through an import map when we assume support for `import:dep` URLs. This `import:` URLs are effectively a prerequisite for such a manifest. Are we sure this syntax will continue to make progress?

[...]

PR #210, opened 2020-02-22
Depcache spec and reference implementation proposal

guybedford:
This PR implements the proposal in https://github.com/WICG/import-maps/issues/209, providing an optimal solution to the waterfall problem, while ensuring that import maps remain the source of truth for the graph and optimization for tools that generate them.

Let's keep proposal feedback to the issue, and direct spec feedback to this PR.

The approach I've implemented here is to effectively integrate the depcache check into `fetch a single module script` in the HTML spec, where the natural recursion between those steps and the preload steps perform the graph traversal, while naturally stopping due to the module map existence checks, avoiding the need for visited lists and such.

To support the new `resolve a module specifier` signature, I've separated out the top-level call to it from HTML from the base-level API-like call to it, called `resolve import maps`.

Any changes to the `fetch a single module script` signature in further refactoring can be applied equally to the depcache preload algorithm, although it is worth noting here that the script for depcache preloads doesn't exist at the time of the network request.

Feedback welcome, happy to keep working on this.

### DIFF
--- a/reference-implementation/__tests__/helpers/common-test-helper.js
+++ b/reference-implementation/__tests__/helpers/common-test-helper.js
@@ -3,6 +3,7 @@ const assert = require('assert');
 const { URL } = require('url');
 const { parseFromString } = require('../../lib/parser.js');
 const { resolve } = require('../../lib/resolver.js');
+const { traceDepcache } = require('../../lib/depcache.js');
 
 function assertNoExtraProperties(object, expectedProperties, description) {
   for (const actualProperty in object) {
@@ -59,7 +60,7 @@ function runTests(j) {
   assertNoExtraProperties(
     j,
     [
-      'expectedResults', 'expectedParsedImportMap',
+      'expectedResults', 'expectedParsedImportMap', 'expectedDepcache',
       'baseURL', 'name', 'parsedImportMap',
       'importMap', 'importMapBaseURL',
       'link', 'details'
@@ -85,8 +86,9 @@ function runTests(j) {
     }
     assert(
       'expectedResults' in j ||
-           'expectedParsedImportMap' in j,
-      'expectedResults or expectedParsedImportMap should exist'
+           'expectedParsedImportMap' in j ||
+           'expectedDepcache' in j,
+      'expectedResults, expectedParsedImportMap or expectedDepcache should exist'
     );
 
     // Resolution tests.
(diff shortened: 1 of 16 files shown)


marcoscaceres:
@guybedford, thanks! We will need you to join the WICG to merge this. You should have an invite in your inbox. Let me know once that is done, and then we can merge this.

hiroshige-g:
Brainstorming some issues (not hard, so I expect we can figure out solutions later; right now I don't have bandwidth though):

I prefer not to call `#preload-a-dependency-graph-cache` from `fetch a single module script`, because:

- In the spec, `fetch a single module script` is, as indicated by the name, only for loading a single module, while the callers of `fetch a single module script` is responsible for traversing module graphs. This separation keeps Chromium implementation cleaner, so I'd like to keep this separation.
- Also, `fetch a single module script` is an internal helper concept in the spec. Embedding the preloading behavior within such an internal helper (and in the form of mutually recursive calls) might be unclear.

Also, I suppose the current PR uses `module map` to avoid infinite loops (i.e. if `moduleMap[urlString]` is already set, we don't preload further), but this is different from how module graph loading avoids infinite loops (e.g. https://html.spec.whatwg.org/multipage/webappapis.html#fetch-the-descendants-of-a-module-script uses `visited set`). I feel having a single source of infinte loop avoidance is preferrable for long-term code health.

guybedford:
@marcoscaceres thanks for the info here. I should already be a member I believe? I was sent an email to link my GitHub profile, which I did after posting this PR, but the check hasn't seemed to have updated. Any tips on further steps I can take here welcome.

domenic:
I think this might be best in a separate repository, so it can gather its own community and interested implementers.

guybedford:
@domenic I'm a little weary of aiming to get interest for a spec on top of a spec on top of a spec. The calls from such a depcache spec to WICG import maps will be quite hand-wavey with some complex interactions. In addition I did need to perform some injection and refactoring to the existing import maps algorithms. If you think this is the best approach, then I think a fork might be preferable, with the intent to re-merge. I would just like to see SystemJS and ES module shims built to spec work here on the optimization problems which is my primary interest as an implementor, so this seems like the right sort of route to me. Would you be ok with that?

domenic:
I mean, I have no objection to you forking. I just think it's important to separate the import maps community in this repository from any additional proposals, which have different levels of support. I'm OK with keeping one issue open (e.g. either #208 or #209) to direct folks from this community into that separate body of work, but I don't think it's a good idea to accept proposals like this into the import maps spec without (browser) implementer interest.

marcoscaceres:
@guybedford thanks! The IPR-checking system has found you :) all green.

guybedford:
@domenic is there any way you could still be convinced this is a useful feature for this group to consider? Have you followed the issue discussion? I just haven't really heard a real argument from you for why you don't think this feature is suitable for this spec. Personally I see this as a critical feature for production modules, and first opened an issue on this problem three years ago. I'm happy to assist with spec work, implementation work and even creating a Chromium CL for it.

domenic:
I'm just looking to finish import maps. I am sure there are a ton of useful features in the modules space, of varying criticality in different peoples' judgments. I think they need to build their own communities and support. (I have moved on from working on modules, so import maps will be my last project in that space.)

littledan:
@hiroshige-g I'm sympathetic with the aim of simplifying/unifying recursion and working with cycles (it's complicated on the JS side too!), but it looks like if this preloading happened during "fetch the descendants of a module script", it would happen after the fetch has actually completed and we have the whole module, which seems unnecessarily delayed. Was that the intention?

marcoscaceres:
Ah, ok! if import-maps is mostly "done" per https://github.com/WICG/import-maps/pull/210#issuecomment-590918055 - and more or less ready to migrate out of the WICG, then yes, I agree with @domenic forking would be appropriate here.

PR #227, opened 2020-10-21
Only apply /-prefix handling to special URL-like specifiers

### LINKED ISSUE #166: data: / blob: URL mapping
Brought up by @jkrems on Twitter - do we permit these mappings? And if so, does that mean that a data: URI with a trailing slash is permitted as a folder mapping?

eg:

```json
{
  "data:application/javascript,console.log('test');//null comment//": "https://site.com/"
}
```

would the above map any content after the data URI above to the local URL? And conversely:

```json
{
  "x/": "data:application/javascript,//"
}
```

would the above map `x/\nconsole.log('hi')` into its executable data URI?

domenic:
Closes #166.

***
<a href="https://pr-preview.s3.amazonaws.com/WICG/import-maps/pull/227.html" title="Last updated on Oct 22, 2020, 3:37 PM UTC (1863452)">Preview</a> | <a href="https://pr-preview.s3.amazonaws.com/WICG/import-maps/227/1443033...1863452.html" title="Last updated on Oct 22, 2020, 3:37 PM UTC (1863452)">Diff</a>

### DIFF
--- a/reference-implementation/__tests__/json/url-specifiers.json
+++ b/reference-implementation/__tests__/json/url-specifiers.json
@@ -9,7 +9,14 @@
       "/test/": "/lib/url-trailing-slash/",
       "./test/": "/lib/url-trailing-slash-dot/",
       "/test": "/lib/test1.mjs",
-      "../test": "/lib/test2.mjs"
+      "../test": "/lib/test2.mjs",
+      "data:text/": "/lib/test-data/",
+      "about:text/": "/lib/test-about/",
+      "blob:text/": "/lib/test-blob/",
+      "blah:text/": "/lib/test-blah/",
+      "file:text/": "/lib/test-file/",
+      "ftp:text/": "/lib/test-ftp/",
+      "ws:text/": "/lib/test-ws/"
     }
   },
   "importMapBaseURL": "https://example.com/app/index.html",
@@ -46,6 +53,24 @@
     "should use the last entry's address when URL-like specifiers parse to the same absolute URL": {
       "expectedResults": {
         "/test": "https://example.com/lib/test2.mjs"
+      }
+    },
+    "Non-special vs. special schemes": {
+      "expectedResults": {
+        "data:text/javascript,console.log('foo')": "data:text/javascript,console.log('foo')",
+        "data:text/": "https://example.com/lib/test-data/",
+        "about:text/foo": "about:text/foo",
+        "about:text/": "https://example.com/lib/test-about/",
+        "blob:text/foo": "blob:text/foo",
+        "blob:text/": "https://example.com/lib/test-blob/",
+        "blah:text/foo": "blah:text/foo",
+        "blah:text/": "https://example.com/lib/test-blah/",
+        "ftp:text/foo": "https://example.com/lib/test-ftp/foo",
+        "ftp:text/": "https://example.com/lib/test-ftp/",
+        "file:text/foo": "https://example.com/lib/test-file/foo",
+        "file:text/": "https://example.com/lib/test-file/",
+        "ws:text/foo": "https://example.com/lib/test-ws/foo",
+        "ws:text/": "https://example.com/lib/test-ws/"
       }
     }
   }
--- a/reference-implementation/__tests__/json/url-specifiers.json
+++ b/reference-implementation/__tests__/json/url-specifiers.json
@@ -9,14 +9,7 @@
       "/test/": "/lib/url-trailing-slash/",
       "./test/": "/lib/url-trailing-slash-dot/",
       "/test": "/lib/test1.mjs",
-      "../test": "/lib/test2.mjs",
-      "data:text/": "/lib/test-data/",
-      "about:text/": "/lib/test-about/",
-      "blob:text/": "/lib/test-blob/",
-      "blah:text/": "/lib/test-blah/",
-      "file:text/": "/lib/test-file/",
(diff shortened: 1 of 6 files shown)


spec.bs
```diff
@@ -397,25 +398,31 @@ To <dfn>register an import map</dfn> given an {{HTMLScriptElement}} |element|:
...
       1. If |resolutionResult| is null, then throw a {{TypeError}} indicating that resolution of |specifierKey| was blocked by a null entry.
         <p class="note">This will terminate the entire [=resolve a module specifier=] algorithm, without any further fallbacks.</p>
       1. Assert: |resolutionResult| is a [=URL=].
       1. Return |resolutionResult|.
-    1. If |specifierKey| ends with U+002F (/) and |normalizedSpecifier| [=string/starts with=] |specifierKey|, then:
+    1. If all of the following are true:
+
+      * |specifierKey| ends with U+002F (/),
+      * |normalizedSpecifier| [=string/starts with=] |specifierKey|, and
+      * either |asURL| is null, or |asURL| [=is special=]
```
guybedford, line 423: When would `asURL` be `null` here if the invariant is that it is already normalized?

guybedford (approved):
The text and restriction seem like a good approach to me.

I just hope we don't restrict custom builtin schemes and alternative schemes though.

For example we might want to include safelisted schemes - https://github.com/whatwg/html/pull/5482.

hiroshige-g (approved):
LGTM

reference-implementation/__tests__/json/url-specifiers.json
```diff
@@ -47,6 +54,24 @@
...
+        "blob:text/": "https://example.com/lib/test-blob/",
+        "blah:text/foo": "blah:text/foo",
+        "blah:text/": "https://example.com/lib/test-blah/",
+        "ftp:text/foo": "https://example.com/lib/test-ftp/foo",
+        "ftp:text/": "https://example.com/lib/test-ftp/",
+        "file:text/foo": "https://example.com/lib/test-file/foo",
+        "file:text/": "https://example.com/lib/test-file/",
+        "ws:text/foo": "https://example.com/lib/test-ws/foo",
+        "ws:text/": "https://example.com/lib/test-ws/"
+      }
```
hiroshige-g, line 74: Could you add HTTP, WSS (and HTTPS) schemes?
HTTPS scheme is tested elsewhere, but HTTP and WSS schemes are not tested at all.

reference-implementation/__tests__/json/url-specifiers.json
```diff
@@ -47,6 +54,24 @@
       "expectedResults": {
         "/test": "https://example.com/lib/test2.mjs"
       }
+    },
+    "Non-special vs. special schemes": {
```
hiroshige-g, line 58: FYI this section currently needs blink internals APIs on WPT+Chromium, as `data:`, `about:` etc. schemes cannot be intercepted by service workers. I'll create a separate PR to move this section to e.g. `url-specifiers-internal.json`.

hiroshige-g: Or you can split this section in this PR out to a separate file like:

```
{
  "importMap": {
    "imports": {
      "data:text/": "/lib/test-data/",
      "about:text/": "/lib/test-about/",
      "blob:text/": "/lib/test-blob/",
      "blah:text/": "/lib/test-blah/",
      "file:text/": "/lib/test-file/",
      "ftp:text/": "/lib/test-ftp/",
      "ws:text/": "/lib/test-ws/"
    }
  },
  "importMapBaseURL": "https://example.com/app/index.html",
  "baseURL": "https://example.com/js/app.mjs",
  "name": "URL-like specifiers",
  "tests": {
    "Non-special vs. special schemes": {
      "expectedResults": {
        "data:text/javascript,console.log('foo')": "data:text/javascript,console.log('foo')",
        "data:text/": "https://example.com/lib/test-data/",
        "about:text/foo": "about:text/foo",
        "about:text/": "https://example.com/lib/test-about/",
        "blob:text/foo": "blob:text/foo",
        "blob:text/": "https://example.com/lib/test-blob/",
        "blah:text/foo": "blah:text/foo",
        "blah:text/": "https://example.com/lib/test-blah/",
        "ftp:text/foo": "https://example.com/lib/test-ftp/foo",
        "ftp:text/": "https://example.com/lib/test-ftp/",
        "file:text/foo": "https://example.com/lib/test-file/foo",
        "file:text/": "https://example.com/lib/test-file/",
        "ws:text/foo": "https://example.com/lib/test-ws/foo",
        "ws:text/": "https://example.com/lib/test-ws/"
      }
    }
  }
}
```

hiroshige-g:
FYI Chromium CL: https://chromium-review.googlesource.com/c/chromium/src/+/2491594

spec.bs
domenic: asURL is null if normalizedSpecifier is something like "moment".

reference-implementation/__tests__/json/url-specifiers.json
domenic: I split it out into a separate file, and included HTTP, HTTPS, and WSS.

domenic:
> I just hope we don't restrict custom builtin schemes and alternative schemes though.
>
> For example we might want to include safelisted schemes - whatwg/html#5482.

Protocol handlers only apply to navigation, not to resource fetching, so I don't think that would be the right layering.

PR #229, opened 2020-10-22
Prevent backtracking above /-prefix entries

### LINKED ISSUE #207: Trailing slash map backtracking
In the Node.js modules group work on exports, which in many ways shares similiarities with import maps resolution, we've explicitly added some validation to ensure that mappings cannot be back tracked - effectively that `import 'mapped/path/../../../backtrack'` will throw when backtracking below the target mapping's trailing slash base path.

For the example case:

```js
{
  "imports": {
    "mapped/path/": "/path/to/map/"
  }
}
```

the validation step is basically just to check that any resolution using this imports, after applying the full URL resolution also goes through a check that `resolved.startsWith(targetPathWithTrailingSlash)`, and if not to throw a validation error.

For Node.js package isolation can be still worked around using absolute URLs, but providing this level of encapsulation makes it more obvious to users that private modules should be private.

Whether that same level of encapsulation encouragement is important to import maps is a question worth considering though, so I thought I would bring it up at least.

domenic:
Closes #207.

Also changes the error message suggestion for a nearby case to be more helpful and accurate.

***
<a href="https://pr-preview.s3.amazonaws.com/WICG/import-maps/pull/229.html" title="Last updated on Dec 2, 2020, 7:05 PM UTC (35de9f4)">Preview</a> | <a href="https://pr-preview.s3.amazonaws.com/WICG/import-maps/229/8be56ee...35de9f4.html" title="Last updated on Dec 2, 2020, 7:05 PM UTC (35de9f4)">Diff</a>

### DIFF
--- a/reference-implementation/__tests__/json/packages-via-trailing-slashes.json
+++ b/reference-implementation/__tests__/json/packages-via-trailing-slashes.json
@@ -6,7 +6,8 @@
       "lodash-dot": "./node_modules/lodash-es/lodash.js",
       "lodash-dot/": "./node_modules/lodash-es/",
       "lodash-dotdot": "../node_modules/lodash-es/lodash.js",
-      "lodash-dotdot/": "../node_modules/lodash-es/"
+      "lodash-dotdot/": "../node_modules/lodash-es/",
+      "mapped/path/": "https://github.com/WICG/import-maps/issues/207/"
     }
   },
   "importMapBaseURL": "https://example.com/app/index.html",
@@ -37,6 +38,17 @@
       "expectedResults": {
         "underscore/": null,
         "underscore/foo": null
+      }
+    },
+    "backtracking via ..": {
+      "expectedResults": {
+        "mapped/path/": "https://github.com/WICG/import-maps/issues/207/",
+        "mapped/path/..": null,
+        "mapped/path/../backtrack": null,
+        "mapped/path/../../backtrack": null,
+        "mapped/path/../../../backtrack": null,
+        "moment/../backtrack": null,
+        "moment/..": null
       }
     }
   }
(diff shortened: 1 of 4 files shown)


guybedford (approved):
Nice to see and this is what we implemented in Node.js as well (in addition to banning node_modules segments in the target).

reference-implementation/__tests__/json/packages-via-trailing-slashes.json
```diff
@@ -6,7 +6,8 @@
       "lodash-dot": "./node_modules/lodash-es/lodash.js",
       "lodash-dot/": "./node_modules/lodash-es/",
       "lodash-dotdot": "../node_modules/lodash-es/lodash.js",
-      "lodash-dotdot/": "../node_modules/lodash-es/"
+      "lodash-dotdot/": "../node_modules/lodash-es/",
+      "mapped/path/": "https://github.com/WICG/import-maps/issues/207/"
```
hiroshige-g, line 10: Could you add an entry of `"mapped/"` here, and `"mapped/backtrack"` to the `expectedResults` below?
This is to test that the backtracking failure at a more specific key doesn't allow fallback to less specific key -- i.e. `"mapped/path/../backtrack"` can't be resolved while `"mapped/backtrack"` can be resolved.

reference-implementation/__tests__/json/packages-via-trailing-slashes.json
```diff
@@ -6,7 +6,8 @@
...
     }
```
hiroshige-g, line 11: Also, how about adding test cases for URL-like specfiers + backtracking? (perhaps in `url-specifiers.json`, where URL-like specifiers + trailing slash is tested)

hiroshige-g (approved):
LGTM, once the comments on tests are addressed.

domenic:
Note to self: PR title is wrong; it should be `/-suffix` not `/-prefix`

Issue #2, opened 2018-03-15
How do we install package name maps? Worker/worklet contexts


domenic:
We have a sketch of an idea for how to supply a package name map to a worker:

```js
new Worker(someURL, { type: "module", packageMap: ... });
```

This is interesting to contrast with the mechanism for window contexts proposed currently (and discussed in #1):

- It's unlike the Window-context mechanism in that the creator of the realm controls the realm's module resolution, instead of the realm itself
- It's like the Window-context mechanism in that the script that runs in the realm does not control the realm's module resolution; the control is external to the script itself.

Basically, there is no clear translation of `<script type="packagemap">` into a worker setting.

If we went with this, presumably we'd do the same for SharedWorker. Service workers would probably use an option to `navigator.serviceWorker.register()`, and have an impact similar to the other options?

For worklets, I guess you'd do something like `CSS.paintWorklet.setModuleMap({ ... })`. Only callable once, of course.

In all cases, it would be nice to make it easy to inherit a package name map. With the current tentative proposal you could do

```js
new Worker(url, {
  type: "module",
  packageMap: JSON.parse(document.querySelector('script[type="packagemap"]').textContent)
});
```

but this is fairly verbose and a bit wasteful (since the browser has already done all the parsing and processing of the map, and you're making it do that over again). At the same time, inheriting by default seems conceptually weird, since workers are a separate realm.

matthewp:
Could the packagemap script have a property on it, `packageMap` or something so you could use: `document.querySelector('#mymap').packageMap` to get it?

justinfagnani:
@matthewp I'd certainly like to see a `.module` property for `<script type=module>` that's a Promise resolving to the module object. If this ends up being a general feature, maybe we just need a `.content` or `.value` property.

domenic:
Yeah, that seems pretty reasonable. I think I like the idea of a general property, with some name or another.

guybedford:
Could we by default just have workers inherit the package map from the parent window?

That seems to cover the 99% scenario.

Isolation doesn't seem to align with the worker boundary, as much as it does concepts of Realms.

I understand the issues for worklets and service workers but I'm' not sure they should be dominating use cases here.

guybedford:
I was referring to the proposal to create isolated realms, but it's a somewhat misleading tangent.

The question really is what is the use case for having a different package map for the worker, that couldn't be shared with the page?

guybedford:
[...]

PR #230, opened 2020-10-22
Remove advisements about future Fetch-based version

domenic:
I think we should set aside that idea, and close #169. It has serious problems with establishing the appropriate base URL, as discussed in previous issues. So we can remove these paragraphs from the spec and avoid distracting readers with the idea that the spec might dramatically change in the future.

This also removes the indication that we modify "fetch a module worker script graph". It's not clear how we'll integrate in the future, but it seems unlikely it would be via this exact mechanism, since that would flip "acquring import maps" to false immediately, before any modules are ever resolved.

***
<a href="https://pr-preview.s3.amazonaws.com/WICG/import-maps/pull/230.html" title="Last updated on Oct 22, 2020, 6:52 PM UTC (90fc013)">Preview</a> | <a href="https://pr-preview.s3.amazonaws.com/WICG/import-maps/230/8be56ee...90fc013.html" title="Last updated on Oct 22, 2020, 6:52 PM UTC (90fc013)">Diff</a>

### DIFF
--- a/spec.bs
+++ b/spec.bs
@@ -212,19 +212,8 @@ Insert a call to [=wait for import maps=] at the beginning of the following HTML
 - [=fetch an import() module script graph=]
 - [=fetch a modulepreload module script graph=]
 - [=fetch an inline module script graph=]
-- [=fetch a module worker script graph=] (using <var ignore>module map settings object</var>)
 
-<div class="advisement">
-  In this draft of the spec, which inserts itself into these HTML concepts, the settings object used here is the |module map settings object|, not |fetch client settings object|, because [=resolve a module specifier=] uses the import map of |module map settings object|. In a potential future version of the import maps infrastructure, which interjects itself at the layer of the Fetch spec in order to support `import:` URLs, we would instead use |fetch client settings object|.
-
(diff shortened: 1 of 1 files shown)


hiroshige-g:
> I think we should set aside that idea, and close #169. It has serious problems with establishing the appropriate base URL, as discussed in previous issues. So we can remove these paragraphs from the spec and avoid distracting readers with the idea that the spec might dramatically change in the future.

Sounds good to me, but just to clarify the context, are there any changes in our mind or technical updates that contributed to this more complete deprecation of fetch-based import maps? Or do we remove this now given no updates for a while?

> This also removes the indication that we modify "fetch a module worker script graph". It's not clear how we'll integrate in the future, but it seems unlikely it would be via this exact mechanism

Sounds good, totally deferring to https://github.com/WICG/import-maps/issues/2.

> since that would flip "acquring import maps" to false immediately, before any modules are ever resolved.

Flipping "acquring import maps" to false before any modules are ever resolved sounds working as intended to me (acquring import maps should be false at the time of specifier resolution). Anyway I agree removing the section.

domenic:
> Sounds good to me, but just to clarify the context, are there any changes in our mind or technical updates that contributed to this more complete deprecation of fetch-based import maps?

Nothing recent, but I think the known issues around establishing an appropriate base URL are pretty serious technical problems. It's _possible_ they're surmountable, but I can't think of how. So currently fetch-based import maps are in my mental category of "have serious problems and would need a lot of work" instead of "natural extension of specifier-based maps".

> Flipping "acquring import maps" to false before any modules are ever resolved sounds working as intended to me (acquring import maps should be false at the time of specifier resolution).

Hmm, yeah, I think I miswrote this commit message. I think I meant "before the worker script is even fetched".

PR #231, opened 2020-10-22
Update community polyfills and tooling list

domenic:
See #146.

PR #233, opened 2020-10-23
Add link to @node-loader/import-maps.

joeldenning:
See https://github.com/node-loader/node-loader-import-maps


domenic:
Thanks! I'll fold this into #231, so we don't have to worry about the IPR bot.

domenic:
Hmm, do you know why https://www.npmjs.com/package/@node-loader/node-loader-import-maps requires a login? I was hoping npm packages could have npm URLs instead of GitHub URLs, but I don't want to require people to login...

domenic:
Oh, nevermind, it's https://www.npmjs.com/package/@node-loader/import-maps

PR #237, opened 2020-12-03
Sync tests in WPT and WICG repository (WICG side)

hiroshige-g:
WPT side: https://github.com/web-platform-tests/wpt/pull/26733
WICG side: https://github.com/WICG/import-maps/pull/237

After these PRs, the data-driven JSON files are identical
in the WPT and WICG repositories.

This PR (WICG side) moves non-HTTP(S) tests into `-internal.json` files.


domenic (approved):

domenic:
Let me see if I can make the test runner download files from WPT instead so humans don't have to do this sync work...

PR #238, opened 2020-12-03
Semi-automatically sync tests with WPT

domenic:
Now WPT serves as the source of truth for the tests. To sync them with this repository, we just update the SHA in the sync-tests.js file.

/cc @hiroshige-g. Feel free to review if you want (especially if you have thoughts about the general approach), although I can also merge this without review. It's just a nice-to-have.

We should delay merging this until https://github.com/web-platform-tests/wpt/pull/26733 lands, and then update the SHA in this PR to the post-merge SHA.


hiroshige-g:
Thanks for automating this, looks good!

PR #242, opened 2021-01-09
Fix rejection of multiple import maps

hiroshige-g:
Previously, we intended to reject second and subsequent import maps by checking whether `node document's pending import map script is non-null`, but this didn't work because after import maps are registered `pending import map script` is cleared.
This PR fixes this by clearing `acquiring import maps` on preparing the first import map, rejecting second and subsequent import maps on preparation with error events.

Alternative considered:
It might be better to set Document's import map to an empty import map on second import map, because it would prevent developers more effectively from depending on the behavior of multiple import maps cases.
However, this was more complicated to specify (e.g. the first import map should be canceled on preparing the second import map and is hard to receive error events).

WPT: https://github.com/web-platform-tests/wpt/pull/27117
Chromium CL: https://chromium-review.googlesource.com/c/chromium/src/+/2618838

***
<a href="https://pr-preview.s3.amazonaws.com/hiroshige-g/import-maps/pull/242.html" title="Last updated on Jan 9, 2021, 12:35 AM UTC (bb85579)">Preview</a> | <a href="https://pr-preview.s3.amazonaws.com/WICG/import-maps/242/9778be3...hiroshige-g:bb85579.html" title="Last updated on Jan 9, 2021, 12:35 AM UTC (bb85579)">Diff</a>

### DIFF
--- a/spec.bs
+++ b/spec.bs
@@ -132,8 +132,11 @@ Inside the <a spec="html">prepare a script</a> algorithm, we make the following 
 - Insert the following step to [=prepare a script=] step 7, under "Determine the script's type as follows:":
   - If the script block's type string is an [=ASCII case-insensitive=] match for the string "`importmap`", <a spec="html">the script's type</a> is "`importmap`".
 - Insert the following step before <a spec="html">prepare a script</a> step 24:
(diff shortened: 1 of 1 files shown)


guybedford (approved):
Looks good!

domenic (approved):
Perfect, thank you!

PR #260, opened 2021-09-10
Update links in "Community polyfills and tooling"

dmail:


domenic:
As far as I can tell these are two separate packages, with different READMEs, neither of which is deprecated.  Should we include both?

dmail:
I haven't marked `@jsenv/node-module-import-map` as deprecated on npm but it is. 
`@jsenv/importmap-node-module` should be used instead.

Issue #245, opened 2021-01-30
base tag and importmap resolution 


timreichen:
How are paths resolved when both `<base>` tag and an importmap are present in an html file? Which is resolved first?

domenic:
Nice question.

How it works is that all URLs in the import map, and all URL-like specifiers, are converted to absolute URLs first, using the `<base>` element. Then, the now-absolute URL-like specifiers are compared to the now-absolute URL-like specifiers in the import map.

Does this make sense? We can also walk through an example, if you have one, or you can check out the tests in https://github.com/web-platform-tests/wpt/tree/master/import-maps/data-driven/resources, e.g. https://github.com/web-platform-tests/wpt/blob/master/import-maps/data-driven/resources/url-specifiers.json is pretty informative.

timreichen:
Thank you for your quick answer. This is what I assumed and what makes most sense to me. Maybe this information could be added to the readme somewhere?

domenic:
Sure, a pull request would be welcome.

PR #272, opened 2022-04-01
docs: adding a section in readme for <base> element interaction

k-j-kim:
Following up on https://github.com/WICG/import-maps/issues/245.

I couldn't find any documentation on import map's expected interaction with `<base>` element, other than the issue reference above.

I wrote up a snippet based on your explanation and interacting with examples, but could use a bit more clarification:
- is any other aspects of the import map other than the right-hand side affected by the `<base>` element?
- is there a plan to support "bare" relative URLs to match existing URL behaviours, for example in `<script src>`? I read your explanation in https://developer.mozilla.org/en-US/docs/Web/HTML/Element/script, but the discrepancy might confuse some people.

apologies if this conversation is better suited for https://github.com/WICG/import-maps/issues/245!

### DIFF
--- a/README.md
+++ b/README.md
@@ -24,6 +24,7 @@ _Or, how to control the behavior of JavaScript imports_
   - [Dynamic import map example](#dynamic-import-map-example)
   - [Scope](#scope)
   - [Interaction with speculative parsing/fetching](#interaction-with-speculative-parsingfetching)
+  - [`<base>` element](#base-element)
 - [Feature detection](#feature-detection)
 - [Alternatives considered](#alternatives-considered)
   - [The Node.js module resolution algorithm](#the-nodejs-module-resolution-algorithm)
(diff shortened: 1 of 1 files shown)


README.md
```diff
@@ -345,6 +345,23 @@ would give the following resolutions:
 |b        |/scope2/scope3/foo.mjs |/b-3.mjs      |
 |c        |/scope2/scope3/foo.mjs |/c-1.mjs      |
 
+#### `<base>` element
+
+When `<base>` element is present in the document, all URLs and URL-like specifiers in the import map are converted to absolute URLs using the `href` from `<base>`.
+
+The right-hand side of the mapping should be absolute URLs starting with `./`. "Bare" relative URLs like `node_modules/moment/src/moment.js` will not work for reasons stated in [[Bare specifiers for JavaScript modules]](#bare-specifiers-for-javascript-modules).
```
domenic, line 352: I don't think this paragraph is very related to this section. I think it is covered earlier in https://github.com/WICG/import-maps#bare-specifiers-for-javascript-modules. So would it be OK to just remove it?

README.md
```diff
@@ -345,6 +345,23 @@ would give the following resolutions:
...
+
+```html
+<base href="https://www.unpkg.com/vue/dist/">
+<script type="importmap">
+{
+  "imports": {
+    "vue": "./vue.runtime.esm.js",
+    // resolves to https://www.unpkg.com/vue/dist/vue.runtime.esm.js
```
domenic, line 360: Since comments are not valid JSON, I think this is a bit confusing for people. Maybe insert something like

```html
<script>
import("vue"); // resolves to https://www.unpkg.com/vue/dist/vue.runtime.esm.js
</script>
```

below?

domenic (changes requested):

README.md
k-j-kim: Hm, I thought it was worthwhile repeating here since script src will support both:
```
<script src="vue.runtime.esm.js">
<script src="./vue.runtime.esm.js">
```
whereas import map will only support the latter. But I agree that it can be considered redundant. I'll remove it.

domenic (commented):
The new text looks great. A couple of things:

- I don't think it belongs under the "Scoping examples" section. Probably under "Import map processing"?
- Please regenerate the table of contents by installing [doctoc](https://www.npmjs.com/package/doctoc) and then running `doctoc README.md`. If this is difficult (e.g. because you don't have Node.js or npm installed) let me know and I can do it.

domenic (approved):
Perfect, thank you!

k-j-kim:
thank you for the review and merge @domenic! 🎉

PR #283, opened 2022-07-06
Link events to their definition in HTML

dontcallmedom:
Helps with cross-referencing


dontcallmedom:
I guess that will fix the cross-referencing too :)

